PR #45, opened 2026-05-05
fix(daemon/server): bound per-socket reassembly buffer to defeat unbounded-growth DoS

RevealUIStudio:
## Summary

- Bounds the per-socket reassembly buffer in `packages/daemon/src/server.ts` so a client streaming bytes without a newline can no longer grow it indefinitely.
- New env-overridable config: `REVDEV_DAEMON_MAX_LINE_BYTES` (default `1_048_576` = 1 MiB) → `DaemonConfig.maxLineBytes`.
- On overflow: emit JSON-RPC `-32700` parse-error (id `null`) → `socket.destroy()` → client must reconnect.
- 4 new e2e-shaped tests against a real socket with a 4 KiB cap (170 daemon total).

Closes the **MEDIUM-severity** §C-2 fix from `pillar-1-stability-2026-05-05.md`.

## Why

Pre-fix, [`server.ts:874-879`](https://github.com/RevealUIStudio/revdev/blob/test/packages/daemon/src/server.ts#L874) accumulated client bytes into a single `buffer` variable until a newline was seen, with no upper bound. A malicious or stuck client streaming bytes (e.g. an attacker writing `'a' * N`, or a buggy client whose framing logic forgot the trailing `\n`) could exhaust daemon memory via a single open socket. Audit identified this as a free reliability win — trivial fix, contained surface.

## Default rationale

1 MiB is generous for any plausible JSON-RPC frame:

- Largest legitimate payloads are `inference.chat` / `inference.generate` prompts.
- Even a ~250 K-token prompt encoded as JSON fits comfortably.
- Coordination calls (`session.*`, `mail.*`, `tasks.*`, etc.) are kilobytes at most.

And tight enough to defeat the unbounded-growth attack — at 1 GBps a malicious client would only get a 1 ms window before the socket is dropped.

## Test plan

- [x] `pnpm --filter "@revdev/daemon" build` clean (39 ms ESM)
- [x] `pnpm --filter "@revdev/daemon" test` → 170/170 pass (4 new in `socket-buffer-bound.test.ts`)
- [x] `pnpm --filter "@revdev/daemon" typecheck` clean
- [x] `pnpm exec biome check packages/daemon` no errors (4 pre-existing warnings in `server.ts:935` area unrelated)
- [ ] CI green on `fix/daemon-socket-buffer-bound`
- [ ] Post-merge fast-forward of `~/suite/revdev` to test branch and re-run daemon test suite (verify-twice)

## Out of scope (queued in audit, separate PRs)

[...]

PR #47, opened 2026-05-05
fix(daemon/server): graceful-shutdown drain — wait for in-flight RPC handlers before db.close()

RevealUIStudio:
## Summary

- Adds a graceful-shutdown drain to `daemon.close()` so in-flight RPC handlers complete before PGlite teardown.
- New module-level counter `_activeHandlerCount`, incremented before each `await handler(...)` and decremented in finally.
- New `cfg.shutdownGracePeriodMs` (default 5 s, env-overridable via `REVDEV_DAEMON_SHUTDOWN_GRACE_MS`).
- 7 new tests via test-only seams (`_setActiveHandlerCountForTest`, `_drainActiveHandlersForTest`) — 189/189 daemon total.

Closes the **MEDIUM** §C-4 fix from `pillar-1-stability-2026-05-05.md`.

## Why

Pre-fix, [`server.ts:1099-1112`](https://github.com/RevealUIStudio/revdev/blob/test/packages/daemon/src/server.ts#L1099) ran `db.close()` immediately after `server.close()`. A handler mid-`await db.query(...)` could throw because PGlite tore down under it. Three real-world hazards:

1. **Long handler observed in §C-3**: `worktree.create` shells out to git on a private base branch — credential prompt, large repo, hung remote can hold the handler 30+ s.
2. **Slow Neon dual-write**: `mail.send`, `tasks.create`, `events.log` all do PGlite + Neon HTTP roundtrip — Neon hang stalls the handler.
3. **`inference.*`** with the §C-1 timeouts can still spend up to 30 minutes (model pull) inside the handler before the abort fires.

After §C-3 introduced the daemon shutdown signal, these handlers all honor it (git child gets SIGTERM, fetch aborts). But the close() implementation didn't wait for them to actually finish — it just aborted-and-tore-down. This PR closes that race.

## Sequence in close()

```
1. _shutdownController?.abort()         ← signals every abort-aware helper
2. server.close()                       ← stop accepting new connections
3. drainActiveHandlers(graceMs)         ← poll _activeHandlerCount until 0 or deadline
4. (if deadline) log.warn(remaining)    ← visible in daemon.log; not hung
5. db.close()                           ← safe — no active queries
6. unlink(socketPath)
```

The drain is almost always a no-op in production because handlers honor the abort. The deadline only fires when a handler is genuinely stuck and ignoring the signal — those cases get logged loudly rather than crashing.

## Test seams (@internal)

```ts
export function _setActiveHandlerCountForTest(n: number): void
export function _drainActiveHandlersForTest(deadlineMs: number, tickMs?: number): Promise<{drained, remaining}>
```

These let the unit tests exercise drain semantics directly — no need for a test-only RPC handler registration or live long-running handlers (Ollama / git hangs would be flaky and slow). The `_*ForTest` naming makes them clearly out-of-bounds for production callers.

## Test plan

- [x] `pnpm --filter "@revdev/daemon" build` clean (37 ms ESM)
- [x] `pnpm --filter "@revdev/daemon" test` → 189/189 (7 new in `shutdown-drain.test.ts`)
- [x] `pnpm --filter "@revdev/daemon" typecheck` clean
- [x] `pnpm exec biome check` no errors (2 pre-existing warnings unrelated to this PR — both `!` non-null assertions in `cli.ts:87` + `server.ts:1025`)
- [ ] CI green
- [ ] Post-merge fast-forward of `~/suite/revdev` to test branch and re-run daemon test suite (verify-twice)

## Out of scope (queued in audit)

| PR | Surface |
|---|---|
| §C-5 | `mail.broadcast` transaction + batched Neon insert |
| §C-6 | internal docs gap-files for `memory.*`/`merge.*` Neon mirror + `file_reservations` TTL drift + `mail.markRead` atomicity |
| §B-1..§B-6 | hook hygiene migrations (`usePollingFetch` helper + cluster migrations) |
| §A-1..§A-3 | CI infra (RevealUI-native retry composite action — owner-redirected from third-party) |

Related: [#43](https://github.com/RevealUIStudio/revdev/pull/43) (Pillar 1 warm-up), [#44](https://github.com/RevealUIStudio/revdev/pull/44) (§C-1 inference timeouts), [#45](https://github.com/RevealUIStudio/revdev/pull/45) (§C-2 socket buffer bound), [#46](https://github.com/RevealUIStudio/revdev/pull/46) (§C-3 git child timeout + shutdown infra). #46 in particular: this PR consumes that infrastructure — the `_shutdownController.abort()` signaled to handlers is what makes the drain almost-always a no-op in production.

### DIFF
--- a/packages/daemon/src/__tests__/shutdown-drain.test.ts
+++ b/packages/daemon/src/__tests__/shutdown-drain.test.ts
@@ -241,4 +241,25 @@ describe('shutdown gate (regression — Codex P2 catch on revdev#47)', () => {
     expect(resp.result?.pong).toBe(true);
     await d2.close();
   });
+
+  it('destroys persistent sockets on close — pre-existing connections cannot dispatch RPCs against a closed daemon (Codex round-2 catch)', async () => {
+    const d = await startDaemon({ socketPath, dataDir });
+
+    // Open a long-lived socket BEFORE close() — like a Studio app would.
+    const sock: Socket = connect(socketPath);
+    await new Promise<void>((resolve, reject) => {
+      sock.once('connect', () => resolve());
+      sock.once('error', reject);
+    });
+    expect(sock.destroyed).toBe(false);
+
+    // Trigger close(). Per the new sequence, all open sockets get
+    // destroyed before the drain.
+    await d.close();
+
+    // The pre-existing socket should now be destroyed. If it weren't,
+    // a `data` event could still fire and dispatch against the closed
+    // PGlite — the very race Codex flagged.
+    expect(sock.destroyed).toBe(true);
+  });
 });
--- a/packages/daemon/src/__tests__/shutdown-drain.test.ts
+++ b/packages/daemon/src/__tests__/shutdown-drain.test.ts
@@ -12,12 +12,14 @@
  */
 
 import { mkdtemp, rm } from 'node:fs/promises';
+import { connect, type Socket } from 'node:net';
 import { tmpdir } from 'node:os';
 import { join } from 'node:path';
 import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
 import {
   _drainActiveHandlersForTest,
   _setActiveHandlerCountForTest,
+  _setClosingForTest,
   startDaemon,
 } from '../server.js';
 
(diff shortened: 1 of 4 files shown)


RevealUIStudio:
@codex review

Addressed P2 in commit 61be7cc — added module-level  boolean set at the START of close() (before drain), checked in dispatch BEFORE the counter increment. Dispatch now responds with JSON-RPC -32099 ("Server is shutting down") and bails for any request arriving on an existing socket once shutdown began. Closes the race where a request could increment the counter past the drain observation window.

2 new regression tests:
- rejects RPCs with -32099 once _closing is set, before incrementing the handler counter
- clears _closing on fresh startDaemon so gate is independent across lifecycles

191/191 daemon tests green.

RevealUIStudio:
@codex review

Round-2 P2 addressed in commit cd7f886. Now tracks every accepted socket in `openSockets` set and force-destroys all of them inside close() BEFORE the drain — `server.close()` alone leaves them alive and their data handlers could fire post-shutdown. The socket close handler is also now gated on `!_closing` so it won't race with `db.close()`.

1 new regression test verifies persistent sockets are actually destroyed by close(). 192/192 daemon tests green.

PR #88, opened 2026-05-24
feat(daemon): license expiry warnings + fail-closed start + KEY_FILE support

RevealUIStudio:
## Summary

Hardens the RevDev daemon's license lifecycle: expiry warnings, fail-closed startup on an expired license, and file-based license loading.

## Changes

- **`REVEALUI_LICENSE_KEY_FILE`** — load the license JWT from a file path (priority: inline `REVEALUI_LICENSE_KEY` > file). Lets a systemd-user unit decrypt the key to tmpfs and shred it post-stop instead of persisting it in the process environment. A file that is set-but-unreadable/empty fails loudly (`LicenseConfigError`).
- **Expiry warnings** — warns at 14d (info) / 7d (warning) / 1d (critical) before expiry, on startup and on a daily re-check while running.
- **Fail-closed start** — a present-but-expired license makes the daemon refuse to start (clean non-zero exit, no stack trace). See the design note below.
- **Telemetry** — `license_valid` + `license_expires_timestamp_seconds` Prometheus gauges; `license.expiry-approaching` / `license.expired` events recorded for the dashboard.
- **`--version`** flag on the daemon CLI.
- **`systemd/license-file.conf.example`** — documented secure file-load drop-in (decrypt to `PrivateTmp`, shred on stop).
- **Structured failure `code`** on the JWT verifier so callers branch on `expired` without string-matching the human-readable reason.

## Design note (reviewer input welcome)

Fail-closed is scoped to a **present-but-expired** license — a security event. An **absent** license intentionally still runs in free/degraded mode (session management only), because that unlicensed path is load-bearing across the guard and the existing test suite. (A literal "fail if neither key is set" would have broken free-tier operation.)

## Verification

- typecheck clean
- `guard` + `flows` + new `license-expiry` suites: **71/71** pass (no regression)
- `shutdown-drain`: **10/10** (the new daily-recheck interval is cleared in `close()`)
- package build OK

18 new test cases cover the expiry thresholds (15d/13d/6d/12h/-1h), the fail-closed throw, file-based load, and the config-error path.

### DIFF
--- a/packages/daemon/src/__tests__/license-expiry.test.ts
+++ b/packages/daemon/src/__tests__/license-expiry.test.ts
@@ -1,0 +1,193 @@
+/**
+ * GAP-184 — daemon license-expiry warn/fail behavior + REVEALUI_LICENSE_KEY_FILE.
+ *
+ * Covers: expiry status buckets (14d/7d/1d), fail-closed on expired, the
+ * file-based key source, and the loud config error when KEY_FILE is broken.
+ */
+
+import { mkdtempSync, rmSync, writeFileSync } from 'node:fs';
+import { tmpdir } from 'node:os';
+import { join } from 'node:path';
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
+import { initLicenseGuard } from '../guard.js';
+import {
+  checkLicense,
+  evaluateLicense,
+  LicenseConfigError,
+  LicenseExpiredError,
+  loadLicenseKey,
+} from '../license.js';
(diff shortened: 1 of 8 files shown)


RevealUIStudio:
Addressed in : the daily license-expiry timer now calls  after the local insert (best-effort), mirroring  /  to  like the  handler — so the fleet/admin dashboard receives them when  is configured. Also fixed the biome organizeImports error (guard.ts) that was failing Quality.

PR #90, opened 2026-05-24
feat(scripts): license rotation script + weekly systemd timer

RevealUIStudio:
## Summary

Adds calendar + emergency license rotation for licenses stored in revvault, plus a weekly systemd-user timer. Pairs with the daemon-side expiry warnings / fail-closed work — rotation stages a replacement key *before* the daemon's fail-closed expiry fires.

## Changes

- **`scripts/rotate-license.ts`** — mints a replacement when the current license is within `--threshold-days` of expiry (default 14); idempotent (no-op outside the window). `--emergency --reason "<why>"` rotates immediately (theft / vault anomaly / key exposure). Appends an audit row to `$LICENSE_ROTATION_LOG` (default `~/.local/share/revealui/license-rotation-log.md`) and optionally POSTs a JSON summary to `$LICENSE_ROTATION_ALERT_WEBHOOK`. Defaults target the founder license but `--vault-path` / `--tier` / `--customer` / `--days` generalize it to any license.
- **`scripts/issue-license.ts`** — exports `issueLicense` / `getPrivateKey` / `revvaultSet` behind an `import.meta` main-guard so the rotate script reuses the exact mint path with zero duplication; the CLI is unchanged when run directly.
- **`systemd/revdev-license-rotation.{service,timer}`** — weekly oneshot (Sun 04:00, `Persistent=true`), wired into `pnpm --filter @revdev/daemon setup:systemd` via an additive block in `install.sh`.

## Design notes

- The audit-log path is **env-configurable**, not hardcoded — the operator points `LICENSE_ROTATION_LOG` at a tracked log if they keep one. (No internal paths baked into this public framework repo.)
- The rotation decision (`decodeLicense` + `shouldRotate`) is exported as pure functions for testability; a perpetual license never calendar-rotates (only `--emergency` does).
- `kid` is recorded as `n/a (single-key)` in the audit row — the column is ready for multi-key (`kid`-based) signing-key rotation when that lands.

## Verification

- biome clean (`biome check`)
- `decodeLicense` + `shouldRotate` logic checks pass (decode of valid/perpetual/garbage JWTs; threshold + emergency + perpetual decision matrix)
- both `rotate-license.ts --help` and `issue-license.ts --help` work (refactor's main-guard intact)
- `install.sh` passes `bash -n`

A live rotation smoke (mint a near-expiry license → run the script → confirm replacement) is owner-runnable — it touches the real signing key + revvault and would rotate an actual license, so it isn't run in CI.

### DIFF
--- a/packages/daemon/systemd/install.sh
+++ b/packages/daemon/systemd/install.sh
@@ -69,3 +69,26 @@ systemctl --user status revdev-daemon --no-pager | head -10 || true
 echo
 echo "Tail logs with: journalctl --user-unit revdev-daemon -f"
 echo "If on WSL and you want survival across logouts: sudo loginctl enable-linger \$(whoami)"
+
+# --- License rotation timer (calendar, weekly) ----------------------------
+# Additive: installs the rotation oneshot + timer using the same NODE_PATH +
+# repo resolution as the daemon unit above. A second daemon-reload is harmless.
+ROTATION_SERVICE_TEMPLATE=$(dirname "$0")/revdev-license-rotation.service
+ROTATION_TIMER_TEMPLATE=$(dirname "$0")/revdev-license-rotation.timer
+ROTATION_SCRIPT=$REPO_ROOT/scripts/rotate-license.ts
+ESCAPED_SCRIPT=$(printf '%s\n' "$ROTATION_SCRIPT" | sed 's/[&"]/\\&/g')
+
+if [ -f "$ROTATION_SERVICE_TEMPLATE" ] && [ -f "$ROTATION_TIMER_TEMPLATE" ]; then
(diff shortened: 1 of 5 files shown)


RevealUIStudio:
Addressed both review findings in `3319fb5`:

- **Command injection (CodeQL "indirect uncontrolled command line" + Codex P1, line 175):** all `revvault` invocations now use `execFileSync` with an argv array instead of `execSync` with a shell-interpolated path — `readCurrentLicense` here, plus `getPrivateKey` and `revvaultSet` in `issue-license.ts`. No shell is spawned, so a `--vault-path` containing shell metacharacters can no longer execute anything.
- **Malformed-JWT silent-skip (Codex P1, line 87):** `decodeLicense` now sets a `malformed` flag, distinct from a valid perpetual license (`exp: null`). A non-emergency run **fails fast (exit 1)** on an unparseable vault entry instead of silently no-op'ing; `--emergency` still force-replaces.

Verified locally: biome clean, `decodeLicense`/`shouldRotate` logic checks (including the malformed-vs-perpetual distinction), and `--help` intact.

PR #93, opened 2026-05-26
Merge main into test and update workflows and scripts

RevealUIStudio:


RevealUIStudio:
Codex P1 (`Verify signature before returning expired`) addressed by [#94](https://github.com/RevealUIStudio/revdev/pull/94) (merged to `test` 2026-05-28). `verifyLicenseJWT` now defers the temporal `now > exp` comparison until after signature + iss + aud + nbf + revocation. Forged tokens with backdated `exp` return `code: 'invalid-signature'` (degrade-to-free) instead of `code: 'expired'` (which `evaluateLicense()` maps to fail-closed startup). Three new tests in `license-expiry.test.ts` cover the forged-expired case. Existing `flows.test.ts` "rejects an expired JWT" fixture updated with the required `iss`/`aud` claims so it exercises the legitimate expired path.

Re-run on this PR should now pick up the fix.

PR #94, opened 2026-05-28
fix(daemon): verify license signature before expiration check

RevealUIStudio:
## What
Reorder `verifyLicenseJWT` so signature verification happens **before** the temporal `now > exp` comparison. The cheap syntactic `exp`-claim-type validation stays where it is.

## Why
Codex P1 on PR #93 (Verify signature before returning `expired`):

> Move the expiration classification behind signature verification, otherwise any syntactically valid JWT with a past `exp` is treated as `code: 'expired'` even when the signature is invalid. In this commit, `evaluateLicense()` maps `expired` to fail-closed startup behavior, so a mistyped/forged token with stale claims now prevents daemon startup instead of falling back to degraded free mode (the documented behavior for invalid signatures).

This is real — verified by reading `packages/daemon/src/license-crypto.ts` (expiry check at the old line ~158 fired before the signature check at the old line ~188) and `packages/daemon/src/license.ts` (`evaluateLicense()` maps `code === 'expired'` to fail-closed via `LicenseExpiredError`).

## Change

`packages/daemon/src/license-crypto.ts`:
- `exp`-claim **type** validation stays early (returns `invalid-claim` for non-numeric exp).
- `now > exp` **temporal** check moved to after signature + iss + aud + nbf + revocation.
- Added inline comments explaining both halves so a future reader doesn't re-flatten them.

`packages/daemon/src/__tests__/license-expiry.test.ts`:
- New `describe('verifyLicenseJWT — signature precedes expiration check ...')`:
  - forged token + past exp → `code: 'invalid-signature'` (proves the fix)
  - forged token + future exp → `code: 'invalid-signature'` (regression)
  - properly-signed expired token → `code: 'expired'` (preserved)

## Validation
- All existing tests use the matching signer/vendor keypair (helper at `test-license-helper.ts`), so they continue to hit the legitimate `'expired'` path.
- New tests directly exercise the forged-expired class that motivates the reorder.
- Local typecheck/test deferred to CI (constrained workstation per fleet practice).

## Out of scope
- The other early-return codes (`invalid-format`, `unsupported-algorithm`, `invalid-tier`, `invalid-claim`) are syntactic and don't leak license state — they stay early. Codex's concern was specifically about `expired` reaching callers ahead of signature verification.

### DIFF
--- a/packages/daemon/src/__tests__/flows.test.ts
+++ b/packages/daemon/src/__tests__/flows.test.ts
@@ -169,8 +169,17 @@ describe('verifyLicenseJWT — negative cases', () => {
 
     const now = Math.floor(Date.now() / 1000);
     const header = Buffer.from(JSON.stringify({ alg: 'EdDSA', typ: 'JWT' })).toString('base64url');
+    // iss/aud required: temporal expiry now runs AFTER signature + iss + aud
+    // verification (see verifyLicenseJWT comments). Without them this would
+    // fall through to 'invalid iss' before the expiry check.
     const payloadB64 = Buffer.from(
-      JSON.stringify({ tier: 'pro', iat: now - 7200, exp: now - 3600 }),
+      JSON.stringify({
+        tier: 'pro',
+        iat: now - 7200,
+        exp: now - 3600,
+        iss: 'https://revealui.com',
+        aud: 'revealui-license',
+      }),
     ).toString('base64url');
     const message = `${header}.${payloadB64}`;
     const sig = sign(null, Buffer.from(message), privateKey).toString('base64url');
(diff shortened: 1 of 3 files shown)


RevealUIStudio:
Codex P2 acknowledged. The `flows.test.ts` "rejects an expired JWT" fixture is updated in commit b75fe5b to include the required `iss`/`aud` claims — the test now exercises what its name promises (an expired JWT, with no other claim defects).

Design note: the temporal expiry check sits after signature/iss/aud/nbf/revocation rather than immediately after signature so each fault returns its most-specific code (signature → iss → aud → nbf → revoked → expired). Codex's just-after-signature placement would also be correct in terms of the fail-closed-on-forged property; this placement just preserves more granular fault reporting.

PR #103, opened 2026-06-12
test(studio): Rust integration tests vs real daemon + CI (A5)

RevealUIStudio:
## What

Closes launch-plan task **A5 — Rust integration tests (Tauri bridge ↔ real daemon)**, the last open W1 agent task besides Console productization.

Two integration binaries under `apps/studio/src-tauri/tests/`, both driving a **real daemon child** (`node packages/daemon/dist/cli.js`) over a Unix socket:

**`harness_integration.rs`** — the Rust JSON-RPC client end to end:
- full free-tier round-trip: `ping` → `ensure_session` (register) → `harness.health` (with the injected `actorAgentId`) → `session.list`
- unreachable socket → error **after the retry ladder actually elapses** (asserted via wall-clock lower bound on the 100ms+500ms backoff)
- hard-kill the daemon, replace it on the same socket path → next call succeeds (fresh-connection-per-call resilience). This deliberately **replaces** the spec'd "daemon starts after 1s succeeds" idea — that test races daemon boot time against a ~600ms retry window and would flake; restart-survival proves the same property deterministically.
- RPC-level errors (unknown method) returned immediately, not retried

**`daemon_ctl_integration.rs`** — lifecycle commands as a customer install resolves them: `REVDEV_DAEMON_BIN` shim that `exec`s node (spawned pid IS the daemon pid), `daemon_status` (down) → `daemon_start` (internally polls ping) → `status` (running + reachable, pid matches) → double-start refused → `daemon_stop` (SIGTERM via pid file) → down. Plus clean failure on a broken binary path. Drop guards SIGTERM leftovers so CI never leaks node processes.

**Scope note (honest):** the daemon license-gates `tasks.*`/`mail.*` (pro+); these tests run it deterministic free-tier (license env stripped in the spawn). Transport semantics — timeout, retry, framing, error mapping — are method-agnostic, which is what this suite exists to prove.

## CI

New `studio-rust-tests.yml`:
- **path-filtered** (`apps/studio/src-tauri/**`, `packages/daemon/**`, `packages/protocol/**`, itself) — compiling `studio_lib` pulls tauri + vendored libgit2/openssl; too heavy to run on every PR
- SHA-pinned actions copied from `studio-release.yml`; `swatinem/rust-cache`; daemon built first; empty `apps/studio/dist` created (tauri-build's codegen requires `frontendDist` to exist)
- `--test-threads=1` — the socket override env (`REVDEV_TEST_SOCKET` / `REVDEV_DAEMON_*`) is process-global
- **deliberately NOT a required check** — path-filtered workflows report SKIPPED on unrelated PRs, and SKIPPED blocks merges when required

`lib.rs` exposes `harness` + `daemon_ctl` as `pub` (integration-test boundary). Tokio added as a dev-dependency (no Cargo.lock delta — already resolved via tauri).

**Verification:** Rust is validated by this PR's `Studio Rust Tests` run (first run is cold-cache, expect ~10–20 min). The 4GB local WSL box cannot compile the tauri graph — per the workstation constraint, CI is the compile arbiter; I did not run cargo locally and this PR should merge only on a green `Rust Integration` job.

## Plan

`docs/PLAN.md`: A5 → ✅ SHIPPED with the scope rationale. W1 agent tasks now: A1 ✅ A2 ✅ A4 ✅ A5 ✅, A3 superseded by W3, A6 (Console) open P3.


RevealUIStudio:
First CI run surfaced a **real production bug** in `daemon_ctl` (exactly what this suite exists for):

`daemon_lifecycle_start_status_stop` failed with `Daemon PID 26690 did not exit within 10s` — but the daemon **had** exited cleanly on the first SIGTERM. `daemon_start` spawns the daemon and drops the `Child` handle without ever `wait()`ing on it, so the exited daemon lingers as a **zombie** of the parent (Studio in production, the test binary here). `is_pid_alive` uses `kill(pid, 0)`, which counts zombies as alive → `daemon_stop` times out reporting a daemon that is already dead. A Studio user clicking Stop would hit the same false error.

Fix (in this PR, `daemon_ctl.rs`): a detached reaper thread `wait()`s on the spawned child, so the zombie is collected the moment it exits and `kill(pid, 0)` correctly returns ESRCH. `harness_integration` (4/4) and `daemon_start_fails_cleanly_when_binary_is_broken` passed on the first run; re-running for the lifecycle test.

PR #173, opened 2026-06-24
fix(daemon): make the Ed25519 signature gate a real barrier (enrollment + per-agent root scoping)

RevealUIStudio:
## Make the Ed25519 signature gate a real barrier (daemon enrollment + per-agent root scoping)

**Security: HIGH.** Closes the self-enrollment / key-takeover hole and the flat-global-roots authorization gap that let any process reaching the `0600` socket sign mutations against any project root. This is the owner's trust-domain decision deferred in commit `24cb306`. **Coordinator review requested — do not self-merge.**

### The vulnerability (audited on `origin/test`)
1. `registerClientIdentity` (server.ts) enrolled **any** client-supplied PEM under **any** agentId with zero proof-of-trust. Worse: supplying a *different* key for an *existing* agentId hit the rotation branch and **superseded the legitimate key** (key-takeover, not just self-enroll).
2. `registeredRoots` (filegit.ts) was a **flat module-level `Set`**, not keyed by agent; the file/git handlers didn't even receive `ctx`. Any registered root was usable by any connection.
3. `project.open` was signature-OPTIONAL and registered the root regardless.

Attack chain: a host process `wsl.exe`-es in as the WSL user → `session.register` self-enrolls its key (exempt) → `project.open` (exempt) registers a root → signs `file.write`/`git.commit` against it. The signature gate passed because the attacker's own key was "registered."

### The fix (A + B + minimum, per the directive)
- **Fix A — provisioned-fingerprint enrollment.** `registerClientIdentity` now rejects (`-32004`) any client key whose fingerprint isn't in a **root-owned** trust anchor (`/etc/revdev/trusted-client-fingerprint`, env-overridable via `REVDEV_DAEMON_TRUSTED_CLIENT_FP`). Root-owned because the threat runs *as the WSL user* and could forge a user-owned file. Fail-closed on a missing/unreadable anchor. Runs on every path — insert, rotation, idempotent re-register — so the key-takeover is closed too. Daemon-MINTED identities (headless hooks) never consult the anchor: their private key is generated in-process and never handed out, so they aren't spoofable by a host process.
- **Fix B — per-agent root scoping.** `registeredRoots` is now `Map<root, agentId>`; `ctx` is threaded into the filegit handlers; `requireRoot` enforces caller ownership. "Not registered" and "owned by another agent" throw the **same** error (no cross-agent existence oracle). Signed mutations + content reads authorize on the **verified signer** (`ctx.agentId`); the signature-optional metadata reads (`git.status`/`listBranches`/`log`) accept an `actorAgentId` fallback (still owner-scoped; they were already optional/metadata-only).
- **Minimum — `project.open` is now signature-REQUIRED** (added to `MUTATING_OR_CONTENT_METHODS`), so the root is recorded under the verified signer rather than a spoofable param. The Rust client mirror `requires_signature()` is flipped in lockstep (cross-language contract test updated).
- **Provisioning bundled (per owner decision):** Studio's WSL `daemon_setup` now writes this install's signing fingerprint to the root-owned anchor via sudo, so enforcement + provisioning land atomically and the legit client keeps working ("one install == one trusted client key").

### Tests (the 3 required + guards)
- **(1)** Unprovisioned key rejected at register (`-32004`) — fresh self-enroll AND key-takeover of an existing agent. (`filegit-enrollment.test.ts`)
- **(2)** Agent A's valid signature cannot mutate a root agent B opened; control proves A can write its own root. (`filegit-enrollment.test.ts`)
- **(3)** Legit Studio client still works end-to-end — `filegit-signed.test.ts` / `filegit-git.test.ts` updated to the new model (trust-anchor fixture + signed `project.open`).
- `signature-gate.test.ts` drift guard updated; `signing.rs` `requires_signature_matches_daemon_set` updated.

### Verification (local)
- Daemon: **406/406** tests pass (run serially — parallel runs hit PGlite cold-init contention, not failures), `pnpm typecheck` clean, gitleaks clean.
- Rust: `cargo test signing::` green (incl. the cross-language contract test); crate compiles on Linux.
- **CI gap:** the `daemon_setup` provisioning write is `#[cfg(not(unix))]`, so Linux CI does not compile it — reviewed against the existing `wsl::run` pattern; needs a Windows/WSL runtime smoke (folds into the P4 install pipeline, same as the #164 follow-up note). Native-unix installs must provision the anchor separately (documented in the unix `daemon_setup`).

### DIFF
--- a/apps/studio/src-tauri/src/daemon_ctl.rs
+++ b/apps/studio/src-tauri/src/daemon_ctl.rs
@@ -436,6 +436,14 @@ pub async fn daemon_setup(app: tauri::AppHandle) -> Result<String, String> {
 
 /// Native Unix: the daemon runs as a local process; no WSL staging is needed.
 /// Register a systemd-user unit with packages/daemon/systemd/install.sh.
+///
+/// NOTE: the client-key enrollment gate still applies on native Unix — a
+/// native-Unix Studio install must provision its signing fingerprint into the
+/// root-owned trust anchor (default `/etc/revdev/trusted-client-fingerprint`,
+/// or REVDEV_DAEMON_TRUSTED_CLIENT_FP) or the daemon rejects its key. This is
+/// left to install.sh / a manual step here because native Unix is the dev/test
+/// surface (tests point the daemon at a fixture anchor); production ships
+/// Windows Studio + WSL daemon, handled by the not(unix) daemon_setup above.
 #[cfg(unix)]
 #[tauri::command]
 pub async fn daemon_setup(_app: tauri::AppHandle) -> Result<String, String> {
--- a/apps/studio/src-tauri/src/daemon_ctl.rs
+++ b/apps/studio/src-tauri/src/daemon_ctl.rs
@@ -398,7 +398,40 @@ pub async fn daemon_setup(app: tauri::AppHandle) -> Result<String, String> {
             String::from_utf8_lossy(&out.stderr).trim()
         ));
     }
-    Ok("RevDev relay installed and daemon enabled in WSL.".to_string())
+
+    // Provision the client trust anchor. The daemon rejects enrollment of any
+    // client key whose fingerprint is not in this ROOT-OWNED file (a host
+    // process running as the WSL user could forge a user-owned file, so the
+    // anchor must be root:root). Write THIS install's Studio signing
+    // fingerprint — "one install == one trusted client key". Overwrites (not
+    // appends) so re-running setup after an identity rotation replaces the old
+    // value rather than accumulating stale keys.
+    let identity = crate::signing::load_or_create_identity()?;
+    let fp = identity.fingerprint;
+    // Defense-in-depth: base58 fingerprints are alphanumeric. Refuse to shell-
+    // interpolate anything else (no injection via a malformed identity file).
+    if fp.is_empty() || !fp.chars().all(|c| c.is_ascii_alphanumeric()) {
+        return Err(format!("refusing to provision a malformed client fingerprint: {fp:?}"));
+    }
+    let provision = format!(
+        "set -e; \
+         sudo mkdir -p /etc/revdev; \
+         printf '%s\\n' '{fp}' | sudo tee /etc/revdev/trusted-client-fingerprint >/dev/null; \
+         sudo chmod 0644 /etc/revdev/trusted-client-fingerprint"
+    );
+    let pout = wsl::run(&["bash", "-lc", &provision]).await?;
+    if !pout.status.success() {
+        return Err(format!(
+            "Relay installed, but provisioning the client trust anchor failed (needs sudo). \
+             The daemon will reject Studio's key until \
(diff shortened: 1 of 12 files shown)


RevealUIStudio:
## REQUEST-CHANGES addressed — ready for re-review

All three HIGH holes fixed on this branch, plus the promoted Windows-ACL item. No self-merge.

**B-1 — optional reads no longer trust a spoofable `actorAgentId`.** `git.status` / `git.listBranches` / `git.log` moved into `MUTATING_OR_CONTENT_METHODS` + `signing.rs requires_signature` (harness auto-signs them) and the `actorAgentId` fallback dropped — they now authorize on the verified signer (`ctx.agentId`) only. Red-before-fix test added: unsigned `git.log`/`git.status` with a victim `actorAgentId` against an un-owned root → `-32003` (`filegit-enrollment.test.ts`). Drift guards updated (daemon `signature-gate.test.ts` + Rust `requires_signature_matches_daemon_set`).

**B-2 — trust anchor enforced root-owned; env override neutralized.** `loadTrustedClientEntries` now reads via `readRootOwnedFile`: opens `O_NOFOLLOW`, fstat `uid==0` + not group/other-writable + regular file, and walks every ancestor dir (root-owned, non-symlink, not group/other-writable). Enforcement is a programmatic config field `trustedAnchorRequireRootOwned` (default `true`, **not env-settable**) — so a WSL-user attacker's `Environment=REVDEV_DAEMON_TRUSTED_CLIENT_FP=/tmp/mine` + `systemctl --user restart` points at a non-root file that fails the check. Tests disable it only via the in-process `startDaemon` config (an env attacker can't reach it).

**B-3 — enrollment bound to (agentId, fingerprint).** Anchor lines are now `agentId:fingerprint`; `registerClientIdentity` rejects unless the exact pair is anchored, so one trusted key can't mint unlimited owning agentIds. The Studio WSL provisioning writes the `agentId:fingerprint` pair (`daemon_ctl.rs`), with `validate_anchor_components` rejecting a `:`/metachar-bearing agentId.

**Promoted — Windows seed-file ACL.** `signing.rs` now strips inherited ACEs and grants only the current user on the seed file on Windows (`icacls /inheritance:r /grant:r`), **fail-closed**: if it can't lock the key, it deletes the file and errors rather than leave an unprotected private key.

**Verified locally:** daemon 430/430 (serial); `cargo test` signing (8) + daemon_ctl (3) + harness_integration (4, incl. `full_rpc_round_trip` against the real cli.js daemon with a sudo-provisioned root-owned pair anchor); typecheck + Biome + gitleaks clean. CI green.

**Tracked, not blockers:** #174 (filter.* post-open write-then-stage, narrowed to an authorized signer), #175 (native-unix anchor provisioning), #176 (Windows/WSL runtime smoke for the `not(unix)` provisioning glow + a `cargo check --target windows` job). The not(unix) provisioning's pure logic is now Linux-unit-tested; only the `wsl::run` glue remains uncompiled by Linux CI.

Issue #176, opened 2026-06-24
Windows/WSL runtime smoke for daemon_setup trust-anchor provisioning


RevealUIStudio:
## Windows/WSL runtime smoke for daemon_setup trust-anchor provisioning

Tracked from #173.

The WSL provisioning write lives in `#[cfg(not(unix))] daemon_setup` (apps/studio/src-tauri/src/daemon_ctl.rs), so **Linux CI never compiles it**. #173 mitigated by extracting the platform-independent logic (`validate_anchor_components`, `build_trust_anchor_provision_script`) into Linux-compiled + unit-tested helpers, so only the `wsl::run(...)` orchestration glue is uncompiled.

Still needed: a real Windows + WSL runtime smoke that exercises `daemon_setup` end-to-end — relay staging, `systemctl --user enable`, and the sudo `printf '%s\n' 'agentId:fp' | sudo tee /etc/revdev/trusted-client-fingerprint`, then a Studio `session.register` that enrolls successfully and a signed `file.write` round-trip. Folds into the P4 install pipeline (same follow-up class as the #164 "relay glue type-checked, not run on Windows" note).

Consider also: a `cargo check --target x86_64-pc-windows-*` CI job to at least type-check the `not(unix)` code path.

RevealUIStudio:
Status as of 2026-07-16: still open. The helpers extracted in #173 remain in place and Linux-tested: validate_anchor_components (apps/studio/src-tauri/src/daemon_ctl.rs:374) and build_trust_anchor_provision_script (daemon_ctl.rs:392), with unit tests in the file's cfg(test) module. The not(unix) daemon_setup (daemon_ctl.rs:411 onward) still holds the WSL orchestration glue that only compiles on Windows. PR CI still runs entirely on ubuntu-latest (.github/workflows/ci.yml and studio-rust-tests.yml have no Windows leg), while release builds compile the Windows path via the windows-latest matrix entry in studio-release.yml. No cargo check job for an x86_64-pc-windows target exists, and no end-to-end Windows plus WSL runtime smoke has been added. The issue remains valid as written.

Issue #175, opened 2026-06-24
Native-unix installs don't provision the client trust anchor


RevealUIStudio:
## Native-unix installs don't provision the client trust anchor

Tracked from #173.

The daemon's client-key enrollment gate (root-owned `/etc/revdev/trusted-client-fingerprint`, entries `agentId:fingerprint`) applies on ALL platforms. The Windows `daemon_setup` provisions it via sudo, but the `#[cfg(unix)] daemon_setup` is a no-op — so a **native-unix Studio** install's key is rejected (the daemon fails closed).

Native-unix is the dev/test surface today (in-process tests pass `trustedAnchorRequireRootOwned: false` via the programmatic `startDaemon` config; the harness integration test sudo-provisions `/etc/revdev`). Production ships Windows Studio + WSL daemon, which IS provisioned. So this is a real but non-urgent gap.

**To do:** add provisioning to the unix `daemon_setup` (or `packages/daemon/systemd/install.sh`) — write `agentId:fingerprint` to the root-owned anchor via sudo, mirroring the WSL path. Documented inline in `daemon_ctl.rs` unix `daemon_setup`.

RevealUIStudio:
Status as of 2026-07-16: still open and accurate. The unix daemon_setup at apps/studio/src-tauri/src/daemon_ctl.rs:496-500 still returns "No WSL setup required on this platform." and does not provision the trust anchor, while the not(unix) path provisions /etc/revdev/trusted-client-fingerprint via build_trust_anchor_provision_script (daemon_ctl.rs:392). packages/daemon/systemd/install.sh does not write the anchor either. The doc comment above the unix daemon_setup records this as a deliberate deferral: native unix is the dev/test surface, and production ships Windows Studio plus WSL daemon, which is provisioned. The enrollment gate continues to fail closed for an unprovisioned native-unix install, so behavior is safe but the install path is incomplete. Unix-side provisioning (in daemon_setup or install.sh) remains planned scope.

Issue #174, opened 2026-06-24
filter.* post-open write-then-stage exec vector (narrowed by #173)


RevealUIStudio:
## filter.* post-open write-then-stage (exec-config TOCTOU, narrowed)

Tracked from the #173 security review (reviewer "also track" item).

`project.open` now refuses a repo whose `.git/config` carries exec-bearing keys (#172), and run-time git ops are hardened (`--no-ext-diff`, `core.hooksPath=/dev/null`, env scrub). But a caller that has already opened a root (now: an **authorized signer** that owns the root, after #173's per-agent scoping) can still write a `.git/config` `filter.<name>.process` / `.clean` / `.smudge` after open and then stage a file that triggers it.

The blast radius is narrowed by #173 (only the root's owning signer can do this, not any socket reacher), but it's still a run-time exec vector that the open-time scan doesn't cover.

**Options to evaluate:**
- Re-scan exec-bearing config keys before each mutating git op (cost per op), or
- Pass `-c filter.<...>=` neutralizers / run git with a pinned `GIT_CONFIG_*` that excludes `filter.*` process hooks (mirror the diff.external/hooks approach), or
- Disallow `git.stageFile` when the work tree has `.gitattributes` filters bound to a process command.

Not a #173 blocker — #173 closes the enrollment/authorization holes; this is defense-in-depth on the already-authorized path.

RevealUIStudio:
Addressed by #177 and #181. file.write / file.delete now refuse any target resolving into <root>/.git/ (assertNotGitInternal), which closes the route this issue relied on to plant a filter driver after open; and git.stageFile / git.discardFile / git.switchBranch / worktree.create re-scan local config and refuse if a filter.* driver is present (assertNoFilterDriver). Both behaviors have red/green coverage over the signed socket. Closing.

PR #178, opened 2026-06-24
fix(daemon): SECURITY follow-up to #173 — B-1/B-2/B-3 + Windows ACL (merged too early)

RevealUIStudio:
## SECURITY follow-up to #173 — the 3 HIGH review holes (merged too early)

⚠️ **#173 was merged at its pre-review head `676676b`, which still contained the three HIGH holes the review flagged** ("do NOT merge on green CI — 3 HIGH holes remain"). The fixes had been pushed to the branch *after* the merge, so they did not land in `test`. **This PR carries those fixes.** Until it merges, `test` has the vulnerable enrollment/authorization surface. **Do not self-merge — coordinator re-review.**

Delta only (everything in #173's accepted surface is already in `test`):

**B-1 — optional reads no longer trust a spoofable `actorAgentId`.** `git.status`/`git.listBranches`/`git.log` are now in `MUTATING_OR_CONTENT_METHODS` + Rust `requires_signature` (harness auto-signs); the `actorAgentId` fallback is gone — they authorize on the verified signer only. Red-before-fix test: unsigned `git.log`/`git.status` with a victim `actorAgentId` → `-32003`. Drift guards updated both sides.

**B-2 — anchor enforced root-owned; env override neutralized.** `readRootOwnedFile`: `O_NOFOLLOW` open + fstat (`uid==0`, regular file, not group/other-writable) + every ancestor dir root-owned/non-symlink/not-writable. `trustedAnchorRequireRootOwned` is a programmatic config field (default `true`, **not env-settable**) — a WSL-user attacker's `Environment=REVDEV_DAEMON_TRUSTED_CLIENT_FP=...` points at a non-root file that fails the check.

**B-3 — enrollment bound to (agentId, fingerprint).** Anchor lines are `agentId:fingerprint`; `registerClientIdentity` rejects unless the exact pair is anchored. WSL provisioning writes the pair (`validate_anchor_components` rejects `:`/metachar agentIds).

**Promoted — Windows seed-file ACL.** `signing.rs` strips inherited ACEs + grants only the current user (`icacls`), fail-closed (deletes the key + errors if it can't lock it).

**Verified:** daemon 430/430 (serial); `cargo test` signing (8) + daemon_ctl (3) + harness_integration (4, incl. `full_rpc_round_trip` against the real daemon with a sudo-provisioned root-owned pair anchor); typecheck + Biome + gitleaks clean.

**Tracked, not blockers:** #174 (filter.* post-open, narrowed to an authorized signer), #175 (native-unix provisioning), #176 (Windows/WSL runtime smoke + a `cargo check --target windows` job for the `not(unix)` glue).

### DIFF
--- a/apps/studio/src-tauri/src/daemon_ctl.rs
+++ b/apps/studio/src-tauri/src/daemon_ctl.rs
@@ -343,6 +343,42 @@ pub async fn daemon_stop() -> Result<(), String> {
 
 // ── First-run setup ───────────────────────────────────────────────────────────
 
+/// Root-owned trust anchor path the daemon reads at client-key enrollment time.
+/// MUST match `DaemonConfig.trustedClientFingerprintPath` in
+/// packages/daemon/src/config.ts.
+#[cfg(any(not(unix), test))]
+const TRUST_ANCHOR_PATH: &str = "/etc/revdev/trusted-client-fingerprint";
+
+/// Validate a client signing fingerprint before it is shell-interpolated into
+/// the provisioning script. base58 (bs58) fingerprints are ASCII-alphanumeric;
+/// refuse anything else so a malformed identity file can never inject shell.
+/// Platform-independent + unit-tested on Linux even though the only caller is
+/// the not(unix) WSL `daemon_setup`.
+#[cfg(any(not(unix), test))]
+fn validate_client_fingerprint(fp: &str) -> Result<(), String> {
+    if fp.is_empty() || !fp.chars().all(|c| c.is_ascii_alphanumeric()) {
+        return Err(format!(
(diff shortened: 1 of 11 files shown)


RevealUIStudio:
Merged under owner direction (overriding the coordinator re-review gate, as with #164). This closes the three HIGH holes that shipped via #173's early merge at `676676b`. Fixes verified before merge: daemon 430/430 (serial), cargo signing(8)+daemon_ctl(3)+harness_integration(4 incl. real-daemon round-trip), typecheck/Biome/gitleaks clean. Follow-ups tracked in #174/#175/#176.

PR #180, opened 2026-06-24
fix(studio): re-lock signing key on load + close create-path TOCTOU (seed-ACL MEDIUM/LOW)

RevealUIStudio:
## Seed-ACL follow-up to #178 (MEDIUM + 2 LOW) — must land before test→main

Addresses the re-review of #178: the seed-ACL fix was **create-only**. `load_or_create_at` locked the key only on the create branch; the `path.exists()` load branch early-returned and never re-applied the lock, so an install whose `studio-identity.json` was written under pre-#173 code (inherited %LOCALAPPDATA% ACLs / unix umask) stayed exposed after upgrade. **Coordinator-gated — no self-merge.**

**MEDIUM — re-lock on load.** Factored the locking into an idempotent, fail-closed `lock_identity_file(path)` and call it on the load path before returning. A legacy file is re-locked on next launch; if it can't be locked, the file is deleted and an error returned (never read/use an unsecurable key).

**LOW-1 — create-path TOCTOU.** The create path now writes an **empty** file, locks it, *then* writes the seed — so the private seed is never briefly on disk under inherited ACLs. (Truncate-in-place preserves the DACL/mode just set.)

**LOW-2 — unix fail-open.** `set_permissions(0o600)` was `let _ =` (fail-open); it's now fail-closed inside `lock_identity_file`, matching the Windows path.

**Test** (`signing.rs`, unix-checkable on Linux CI): create → assert `0600`; broaden to `0644` (simulate a pre-#173 file); load again → assert re-locked to `0600` **and** the identity is unchanged (no regeneration). The Windows DACL assertion is covered by the runtime smoke tracked in #176.

**Verified:** `cargo test signing::` 9/9 (incl. the new re-lock test) + `harness_integration` 4/4 (incl. real-daemon round-trip, which loads the identity); gitleaks clean. No TS touched.

### DIFF
--- a/apps/studio/src-tauri/src/signing.rs
+++ b/apps/studio/src-tauri/src/signing.rs
@@ -334,6 +334,34 @@ mod tests {
     use ed25519_dalek::{Verifier, VerifyingKey};
     use serde_json::json;
 
+    #[cfg(unix)]
+    #[test]
+    fn load_relocks_a_world_readable_legacy_identity() {
+        use std::os::unix::fs::PermissionsExt;
+        let dir = std::env::temp_dir().join(format!("revdev-seedacl-{}", std::process::id()));
+        let _ = std::fs::create_dir_all(&dir);
+        let path = dir.join("studio-identity.json");
+        let _ = std::fs::remove_file(&path);
+
+        let mode = |p: &std::path::Path| std::fs::metadata(p).unwrap().permissions().mode() & 0o777;
+
+        // Fresh create → owner-only (0600).
+        let created = load_or_create_at(&path).expect("create identity");
+        assert_eq!(mode(&path), 0o600, "a freshly created signing key must be 0600");
+
+        // Simulate a pre-#173 file: broaden to group/other-readable.
+        std::fs::set_permissions(&path, std::fs::Permissions::from_mode(0o644)).unwrap();
+        assert_eq!(mode(&path), 0o644);
+
+        // Load again → must RE-LOCK to 0600 without regenerating the identity.
+        let loaded = load_or_create_at(&path).expect("load identity");
+        assert_eq!(mode(&path), 0o600, "load must re-lock a legacy key to 0600");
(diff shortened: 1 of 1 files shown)


RevealUIStudio:
@coordinator — merged under owner direction (`62c56ea`). This is the seed-ACL MEDIUM+2LOW follow-up you flagged on the #178 re-review.

**Requesting your post-merge review on `test`** before the `test→main` promotion (you noted these "must land before test→main"):

- **MEDIUM** — `lock_identity_file()` now idempotent + fail-closed and called on BOTH the load and create paths, so a pre-#173 identity file (inherited %LOCALAPPDATA% ACL) is re-locked on next load (delete + Err if unlockable).
- **LOW-1** — create path is now empty-file → lock → write-seed (no seed on disk under inherited ACLs).
- **LOW-2** — unix `chmod 0600` is fail-closed (was `let _ =`).
- Test: create→0600, broaden to 0644, load→re-locked 0600 with identity unchanged (unix, Linux-CI). Windows DACL assertion tracked in #176.

Verified before merge: `cargo test signing::` 9/9 (incl. the new re-lock test) + `harness_integration` 4/4 (real-daemon round-trip) + gitleaks clean; CI green/CLEAN.

**State of the zero-9P security thread on `test`:** #173 (gate) + #178 (B-1/B-2/B-3) + #180 (seed-ACL) all landed. B-1/B-2/B-3 you already verified CLOSED. Open follow-ups tracked, non-blocking: #174 (filter.* post-open), #175 (native-unix provisioning), #176 (Windows/WSL runtime smoke + `cargo check --target windows`).

**Containment note (audited):** `test` held the pre-review #173 head `676676b` from 09:55Z→17:56Z; verified it never reached `main` and no promotion/release/deploy/tag fired in that window — no downstream remediation needed.

If this passes your review, `test→main` is clear for the zero-9P security surface (owner-gated). — agent (WSL-native)

PR #233, opened 2026-07-04
chore(docs): remove an internal engineering audit inadvertently tracked in the public repo

RevealUIStudio:
## What

Removes an internal engineering audit document that was inadvertently tracked in this public repository. It is internal working material and does not belong on the public surface.

## Why

Surfaced by an internal fleet documentation audit (2026-07-03). The file is internal-only content that should never have been committed to a public repo.

## Note for owner — git history

Removing the file from `HEAD` does **not** remove it from the repository's pushed git history — the blob remains retrievable from prior commits and via the GitHub API/clones. Please decide whether a **history purge + force-push** is warranted here (precedent: the 2026-07-02 revkit tarball history purge). If so, that is a separate, owner-run operation after this PR merges.

## Verification

- `docs/audits/` is now empty (the directory held only this one file).
- No code or references point at the removed path.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #234, opened 2026-07-04
docs: reconcile studio release status + macOS caveat, scrub private-key store paths from KEY_GENERATION

RevealUIStudio:
## What

Doc-currency + hygiene pass on three docs, surfaced by an internal audit (2026-07-03):

**`docs/KEY_GENERATION.md`** — the runbook named concrete revvault store paths for the binary/code-signing **private** keys. Genericized those to placeholders and pointers to the internal key index, so the public runbook no longer discloses exact private-key store locations. Public-key wiring steps are unchanged. No secret values were ever present or added.

**`apps/studio/README.md`** — the status line claimed "prebuilt binary release is **not** yet shipped." Reconciled to reality: prebuilt binaries are published on GitHub Releases (first release `studio-v0.1.0`, 2026-07-02; macOS · Linux · Windows). Documented the macOS **unsigned / ad-hoc** first-launch caveat (right-click → Open). In-app auto-update is described as **configured but not yet live** — verified the `releases.revealui.com` update endpoint returns 404 for the update manifests, so it is not presented as working.

**`docs/GETTING_STARTED.md`** — added the same macOS ad-hoc first-launch caveat next to the download table, and the note that in-app auto-update is not live yet. (The release link and the four-tier Free/Pro/Max/Enterprise list were already current.)

## Verification

- `studio-v0.1.0` confirmed published (2026-07-02) with macOS/Linux/Windows assets; `studio-v0.1.1` is now the latest.
- Auto-update endpoint probed: `releases.revealui.com/studio/<target>/<arch>/latest.json` → HTTP 404 (not serving manifests) → documented as "not yet live," not as working.
- No private-key revvault store paths remain in `docs/KEY_GENERATION.md`.

File-disjoint from the companion PR that removes an internal audit doc (#233).

### DIFF
--- a/apps/studio/README.md
+++ b/apps/studio/README.md
@@ -1,6 +1,6 @@
 # RevealUI Studio
 
-> **Status:** Functional in dev — connects to the harness daemon, 506 frontend tests passing. Prebuilt binary release is **not** yet shipped (Tauri `studio-release.yml` CI in progress). Run locally via `pnpm tauri:dev` for now.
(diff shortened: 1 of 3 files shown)


RevealUIStudio:
Monitor review: APPROVED — release-status text verified against tonight's actual state (binaries published, macOS unsigned/ad-hoc caveat matches the CHANGELOG, auto-update endpoint not live). The 'internal key index' pointer verified to resolve: the vault paths are documented in the internal secret-path lane materials, and the in-flight canonical SECRET_PATHS spec is their durable home. Non-blocking note: the same path strings remain in public tooling source (the issuer script), so this scrub reduces casual discoverability rather than eliminating the strings — acceptable and consistent with keeping the tooling functional.

Issue #109, opened 2026-06-12
Implement DID + Ed25519 per-RPC signing in the daemon


RevealUIStudio:
Tracking issue for ADR 2026-05-16-revdev-did-ed25519-rpc-signing (accepted 2026-06-11).\n\nThe daemon currently binds agent identity at socket-connect time with no per-call cryptographic verification. The ADR specifies the full design.\n\nWork items:\n- PGlite schema: agent_identity, agent_identity_keys, agent_identity_nonces tables\n- Key generation: Ed25519 keypair on session.register; stored in revvault at revdev/agents/<agentId>/identity/ed25519-{private,public}\n- DID format: did:revfleet:<agentId>:<fingerprint> (base58btc(sha256(pubkeyBytes)))\n- Signature envelope: method + params hash + nonce + timestamp + kid signed with node:crypto only\n- Verification middleware for all non-IDENTITY_EXEMPT RPCs\n- Rollout: opt-in phase then mandatory phase\n\nAcceptance: all non-exempt RPCs carry valid Ed25519 signatures; nonce replay prevented; zero new external crypto deps; tests for sign/verify, replay, rotation grace period.

RevealUIStudio:
**Codebase status (2026-06-13): opt-in phase implemented; mandatory enforcement remains.**

The ADR 2026-05-16 design is substantially landed in `packages/daemon`:

**Shipped**
- Schema: `agent_identity`, `agent_identity_keys`, `agent_identity_nonces` tables (`src/storage/schema.ts`; migration verified by `migrate.test.ts`).
- Crypto: `src/agent-identity-crypto.ts` — Ed25519 keygen, `signEnvelope`/`verifyEnvelope`, `canonicalizeJSON`, `hashParams`, base64url — **`node:crypto` only, zero external crypto deps**.
- DID: `did:revfleet:<agentId>:<fingerprint>` format + parser (`src/did.ts`).
- Verification middleware in `src/server.ts`: signature parse → DID parse → key lookup (`agent_identity_keys.public_key_pem`) → `verifyEnvelope` → timestamp window → method match → paramsHash match → **nonce replay block** (`INSERT INTO agent_identity_nonces`, conflict → reject).
- `boundVia: 'signature'` identity with a pre-signature snapshot restore (prevents a signed→unsigned reuse from inheriting identity).
- Key rotation via `agent_identity_keys.superseded_at` (exercised in `coordination.test.ts`).
[...]

PR #236, opened 2026-07-04
feat(daemon): P2 signature-status telemetry (observe-only)

RevealUIStudio:
## Summary

Phase P2 of the signature-enforcement rollout (#109): **observe-only signature-status telemetry** in the daemon dispatch loop.

- `emitSignatureTelemetry` emits one `identity.signature_status` event per **non-exempt** RPC, recording `verification` (`verified` / `none` / `invalid`), the method, and whether the method already hard-requires a signature (`required` = membership in `MUTATING_OR_CONTENT_METHODS`).
- Telemetry fires **before** any accept/reject gate, so `none`/`invalid` coverage is measured across all non-exempt traffic — this feeds the P3 flip acceptance gate (≥95% signed over 7d).
- Fire-and-forget: the INSERT is not awaited and failures are swallowed into a debug log — telemetry can never add latency to, throw into, or fail the dispatch path. No accept/reject behavior changes.
- Unauthenticated calls with no `actorAgentId` record the `'unbound'` sentinel (the `agent_id` column is NOT NULL).
- `runPrune` now also deletes `identity.signature_status` rows older than `hardDeleteDays` (same threshold as the existing prune), bounding telemetry-row growth on the periodic pass.

## Tests

New socket-level integration suite `packages/daemon/src/__tests__/signature-telemetry.test.ts` (6 tests) drives a real daemon over a Unix socket and reads emitted rows from the events table: verified / none / invalid statuses, `required=false` for non-mutating methods, `IDENTITY_EXEMPT` silence (`ping`), and prune of stale rows.

## Verification

- `signature-telemetry.test.ts`: 6/6 green
- Full `@revdev/daemon` suite: 27 files, 541 tests green
- `pnpm typecheck` (protocol + daemon + bridge): clean
- Biome: clean on changed files

Refs #109 (ADR 2026-05-16 §Q5 Phase P2). Manual merge after CI green — no `--auto`.

### DIFF
--- a/packages/daemon/src/__tests__/signature-telemetry.test.ts
+++ b/packages/daemon/src/__tests__/signature-telemetry.test.ts
@@ -1,0 +1,269 @@
+/**
+ * P2 signature-status telemetry (ADR 2026-05-16 §Q5).
+ *
+ * The daemon emits one `identity.signature_status` event per NON-EXEMPT RPC
+ * recording whether the call carried a fully-verified Ed25519 signature
+ * (`verified` / `none` / `invalid`) and whether the method already hard-requires
+ * a signature (`required`). This is observe-only — it must change no accept /
+ * reject behavior — and feeds the P3 flip acceptance gate (≥95% signed / 7d).
+ *
+ * These tests drive a real daemon over a Unix socket and read the emitted rows
+ * straight from the events table. The RPC result itself is irrelevant here
+ * (some calls are rejected downstream); telemetry is emitted BEFORE any gate,
+ * so a rejected call still records its signature status.
+ *
+ * @vitest-environment node
+ */
+
+import { mkdtemp, rm, writeFile } from 'node:fs/promises';
+import { connect, type Socket } from 'node:net';
+import { tmpdir } from 'node:os';
+import { join } from 'node:path';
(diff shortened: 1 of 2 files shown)


RevealUIStudio:
Monitor review: APPROVED — sec-review verdict recorded. Verified against the diff: (1) genuinely observe-only — +58/-0, no existing accept/reject line touched; (2) fire-and-forget proven in code (un-awaited INSERT, floating catch to debug log — cannot add latency, throw, or fail dispatch); (3) pre-gate placement so none/invalid coverage measures ALL non-exempt traffic, which is what makes the P3 >=95% flip gate honest; (4) consumes the existing VerificationResult — no new crypto surface; (5) payload is method/status/required only — no signature bytes or key material at rest; (6) prune bounded to the existing hardDeleteDays. Socket-level tests cover all three statuses + exempt silence + prune. One non-blocking flag for P3 prep: the params.actorAgentId fallback is attacker-controllable on unauthenticated calls — an unsigned caller can attribute rows to an arbitrary agent string. Aggregate coverage is unaffected, but tag params-sourced attribution distinctly (e.g. unbound-claimed:<id>) before any PER-AGENT number feeds the flip decision.

PR #237, opened 2026-07-04
ci: add issue/PR body leak-scan gate

RevealUIStudio:
Ports the issue/PR/comment body leak-scan gate from `RevealUIStudio/revealui` to this repo, byte-identical:

- `.github/workflows/issue-leak-scan.yml` — runs gitleaks against issue bodies (`opened`/`edited`), PR bodies (`pull_request_target` `opened`/`edited`), and comment bodies (`created`/`edited`). Pinned gitleaks 8.30.0 installed with SHA256 verification; bot-actor guard prevents loops on the workflow's own remediation comments.
- `.gitleaks.issues.toml` — the issue-surface rule set (extends gitleaks defaults with Stripe object-ID correlation, internal subdomains, private/internal path references, Sentry DSNs, credentialed DB connection strings, EVM wallet addresses, personal-email PII, and confidential-name rules).

On findings the workflow posts a remediation comment (rotate-before-redact guidance), applies the `contains-sensitive-content` label (created in this repo to match revealui's), and fails the run.

Why: this repo is public, so its issue/PR/comment surfaces need the same gate revealui already has. Tracked-content scanning (`scripts/check-no-private-leaks.sh` in CI) is unchanged — this closes the GitHub-conversation-surface gap only.

Note: GitHub runs `issues`/`issue_comment`-triggered workflows from the default branch, so the gate becomes fully active once this rides the next test-to-main promotion. The `pull_request_target` leg activates as soon as the file exists on the base branch.

Manual merge, no auto-merge.

PR #238, opened 2026-07-04
ci: add client-name leak scan for tracked content

RevealUIStudio:
Ports the tracked-content client/prospect-name leak scanner from RevealUIStudio/revealui (`scripts/check-client-leaks.sh` + the "Client Name Leak Scan" workflow) to this repo. Both files are byte-identical to the revealui copies.

- Scans all tracked content on push/PR to `test` + `main`; any hit fails the job.
- The pattern list lives in `scripts/check-client-leaks.sh` (`PATTERNS`); extend it in the same PR that introduces a new name (see the header comment).
- Deliberately has no `.leakignore` escape hatch — the property is unconditional. The scanner self-excludes its own file and the companion `.gitleaks.issues.toml` (issue/PR-body scan, #237), so the two gates never flag each other regardless of merge order.
- File-disjoint from #237 (which owns `.gitleaks.issues.toml`, `issue-leak-scan.yml`, `.leakignore`).
- Verified locally on this branch: `bash scripts/check-client-leaks.sh` exits 0 (repo content is clean today) and `bash scripts/check-no-private-leaks.sh` exits 0 (the ported files trip none of the existing private-leak patterns, so no `.leakignore` entries are needed).
- Follow-up (owner action): mark the "Client Name Leak Scan" workflow as a required status check on `test` + `main`, matching revealui.

### DIFF
--- a/.github/workflows/check-client-leaks.yml
+++ b/.github/workflows/check-client-leaks.yml
@@ -1,0 +1,36 @@
+name: Client Name Leak Scan
+
+# REQUIRED status check on test + main.
+#
+# This workflow hard-fails when any RevealUI Studio client / prospect /
+# warm-intro contact name appears in tracked repo content. Patterns live
+# in scripts/check-client-leaks.sh and are extended in the same PR that
+# introduces a new client into the operator's pipeline.
+#
+# Companion: .github/workflows/issue-leak-scan.yml covers issue + PR body
+# + comment surfaces with the same pattern class (config in
+# .gitleaks.issues.toml). Together the two scanners give blanket coverage
+# across tracked files + GitHub conversation surfaces.
+
+on:
+  push:
+    branches: [test, main]
+  pull_request:
+    branches: [test, main]
+
+permissions:
+  contents: read
+
+jobs:
+  scan:
+    name: Client / prospect name leak scan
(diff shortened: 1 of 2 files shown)


RevealUIStudio:
Post-merge review (recorded for the audit trail): **verified clean.**

- Both added files diff **byte-identical** to the revealui originals (`scripts/check-client-leaks.sh`, `.github/workflows/check-client-leaks.yml`).
- The scanner is path-portable (computes repo root from `BASH_SOURCE`, no hardcoded revealui paths), uses `grep -rFIn` fixed-string matching (no authored regex), and self-excludes its own file plus the companion `.gitleaks.issues.toml`, so it and the issue-body gate never flag each other regardless of merge order.
- Workflow is least-privilege (`contents: read`), pins `actions/checkout` by SHA, `persist-credentials: false`.
- CI ran the scanner against this repo's tracked content: clean (exit 0). Confirmed present and running on `test` post-merge.

No changes needed. One low-priority upstream note (applies to the revealui original, not this port): the matcher is case-sensitive, so only the literal casings listed are caught. Fine for natural-casing accidental leaks; if you want defense against odd casing, harden it in revealui first, then re-port.

PR #259, opened 2026-07-10
fix(studio): daily-driver hardening (SSH hang, orphaned agents, destructive confirms, honest failures)

RevealUIStudio:
Makes Studio safe to use as a daily driver. Every item below was re-verified against current code before being changed; an internal audit had flagged a larger set, but most of it was already remediated and only these remained.

## Crashers and hangs

- **SSH terminal input hang.** The per-session output task held the channel mutex across `ch.wait().await`, so a session parked at an idle prompt pinned the lock and `ssh_send`/`ssh_resize` could never acquire it: typing simply stalled until the server spoke. The channel is now `split()` into a reader-owned `ChannelReadHalf` and a shared `Arc<ChannelWriteHalf>`. `data()`/`window_change()` take `&self`, so no lock is needed at all. The hang is now structurally impossible rather than narrowed.
- **`ssh_send`/`ssh_resize` also held the sessions map lock across their awaits**, so one slow send stalled every other session. They now clone the write half and release the map before awaiting.
- **Orphaned agent processes.** `AgentProcess.child` is a `std::process::Child`, which does not kill on drop, and the exit path reaches `std::process::exit`, which does not run destructors. Quitting Studio left long-lived children (notably `ollama run`) alive. Added `spawner::kill_all()`, invoked on `RunEvent::Exit`.
- **Daemon Stop could fail with no recourse.** `daemon_stop` sent only SIGTERM and, after a 10s window, returned an error while leaving a wedged daemon running. It now escalates to SIGKILL.
- **Tray icon.** `default_window_icon().unwrap()` would panic the app at startup in any build without a bundled icon. Now matched, with a warning.

## Destructive actions without confirmation

Routed the last two unguarded surfaces through the existing `ConfirmDialog` primitive (the other eight already used it):

- `SpawnerPanel` Stop / Remove agent.
- `ConnectForm` SSH bookmark remove.

## Honest failure surfaces

- **Silent token-refresh failure.** A failed refresh was swallowed, so a dead token surfaced later as opaque 401s. A rejected token (`HttpError.kind === 'client'`) now signs the user out with an actionable message. Network and server failures still preserve the deliberate offline-access path. `refreshToken` now returns the error category so callers can tell these apart.
- **`httpRpc` had no timeout**, so a hung remote daemon hung the UI forever. Added a 30s `AbortController` timeout.
- **Browser mode reported a fabricated deployment as real.** `vercelDeploy`/`vercelGetDeployment` now call `markDegraded(...)` and return an obviously fake `MOCK_DEPLOY_ID_DO_NOT_USE` id, matching the existing mock-secret sentinel style.
- **`config.ts` silently discarded settings on reload** in browser mode. It now marks degraded so the shell banner explains it.

Also removed em dashes from the degraded-mode and auth strings these changes touch, per the copy convention.

## Verification

- `cargo check` clean; `cargo test --lib` 83 passed (includes the `wait_for_exit_does_not_block_stop` regression test).
- Rust integration tests via the CI invocation (`--test harness_integration --test daemon_ctl_integration -- --test-threads=1`): 6 passed.
- `pnpm typecheck:studio` clean; `pnpm lint` clean (1 pre-existing warning, untouched file).
- `pnpm --filter studio test`: 541 passed across 63 files. `deploy.test.ts` updated for the new mock deploy id.

## Review note

This touches the auth token lifecycle (`use-auth.ts`, `auth-api.ts`) and the agent-spawn surface (`spawner.rs`). The pre-PR sensitivity self-check reports no coordinator gate, but given those surfaces I have not self-merged. The auth change is fail-closed: a rejected token now de-authenticates where it previously stayed "authenticated."

Windows `platform/windows.rs` still hardcodes the retired `E:\repos` path and runs `reset --hard` against it. It is dead code on Linux and macOS (`create_platform()` only builds `WindowsPlatform` under `cfg(target_os = "windows")`), so it is left for a separate change rather than folded in here.

### DIFF
--- a/.github/workflows/studio-rust-tests.yml
+++ b/.github/workflows/studio-rust-tests.yml
@@ -88,6 +88,7 @@ jobs:
             --manifest-path apps/studio/src-tauri/Cargo.toml \
             --test harness_integration \
             --test daemon_ctl_integration \
+            --test ssh_output_pump \
             -- --test-threads=1
 
   relay:
--- a/apps/studio/src-tauri/src/commands/ssh.rs
+++ b/apps/studio/src-tauri/src/commands/ssh.rs
@@ -58,15 +58,16 @@ pub async fn ssh_send(
         .decode(&data)
         .map_err(|e| StudioError::Other(format!("Invalid base64: {e}")))?;
 
-    let sessions = state.sessions.lock().await;
-    let session = sessions
-        .get(&session_id)
-        .ok_or_else(|| StudioError::Other(format!("No session with id {session_id}")))?;
-
-    let channel_guard = session.channel.lock().await;
-    let channel = channel_guard
-        .as_ref()
-        .ok_or_else(|| StudioError::Ssh("Channel closed".into()))?;
+    // Clone the write half and release the sessions map before awaiting, so a
+    // slow send on one session cannot stall every other session's commands.
+    let channel = {
+        let sessions = state.sessions.lock().await;
+        sessions
+            .get(&session_id)
+            .ok_or_else(|| StudioError::Other(format!("No session with id {session_id}")))?
+            .channel
+            .clone()
+    };
 
     channel
         .data(&bytes[..])
@@ -84,15 +85,14 @@ pub async fn ssh_resize(
     rows: u32,
     state: State<'_, SshState>,
 ) -> Result<(), StudioError> {
-    let sessions = state.sessions.lock().await;
-    let session = sessions
-        .get(&session_id)
-        .ok_or_else(|| StudioError::Other(format!("No session with id {session_id}")))?;
-
-    let channel_guard = session.channel.lock().await;
-    let channel = channel_guard
-        .as_ref()
-        .ok_or_else(|| StudioError::Ssh("Channel closed".into()))?;
+    let channel = {
+        let sessions = state.sessions.lock().await;
+        sessions
+            .get(&session_id)
+            .ok_or_else(|| StudioError::Other(format!("No session with id {session_id}")))?
+            .channel
+            .clone()
+    };
 
     channel
         .window_change(cols, rows, 0, 0)
--- a/apps/studio/src-tauri/src/daemon_ctl.rs
+++ b/apps/studio/src-tauri/src/daemon_ctl.rs
@@ -322,7 +322,22 @@ pub async fn daemon_stop() -> Result<(), String> {
         }
     }
 
-    Err(format!("Daemon PID {pid} did not exit within 10s"))
+    // Escalate. Without this, Stop reports an error while leaving a wedged
+    // daemon running, and the UI offers no way to recover short of a shell.
+    if unsafe { libc::kill(pid as i32, libc::SIGKILL) } != 0 {
+        return Err(format!(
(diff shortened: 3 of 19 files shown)


RevealUIStudio:
**Independent adversarial security review (Fable pass; reviewer session did not author this PR).**

Read the full diff plus the unchanged code the changes lean on (`ssh_disconnect`, `httpRpc` tail, spawner state wiring). Findings by surface:

**SSH channel restructure — sound, and structurally better than a narrower fix.** The `split()` into an exclusively-owned read half (the pump) and an `Arc<ChannelWriteHalf>` removes the lock entirely rather than shrinking its hold window; `data()`/`window_change()` on `&self` cannot re-create the hang class. `ssh_send`/`ssh_resize` correctly clone the write half inside the map-lock scope and await outside it. Teardown verified: user disconnect goes through `handle.disconnect(...)` at the transport level, which ends the pump's `wait()` and runs its map cleanup; a send racing session removal surfaces as a channel error rather than a lock-up. The extracted `pump_output` with the by-value read half plus the do-not-relock comment is the right way to pin the invariant, and the new integration test file covers it.

**`spawner::kill_all` on `RunEvent::Exit` — correct and appropriately scoped.** It only touches `Child` handles the spawner itself owns (no PID sweeping), tolerates a poisoned lock by skipping (the process is exiting), and deliberately skips `wait()` to avoid deadlocking against the reaper thread. Note `std::process::Child::kill` is SIGKILL, so agents get no graceful-shutdown window at app exit; acceptable for this surface, worth remembering if agents ever hold write-ahead state.

**`daemon_stop` SIGKILL escalation — accept.** Theoretical PID-reuse window between the liveness polls and the `kill(2)`, but the PID was the live daemon seconds earlier within the same call; this is the standard pidfile pattern and strictly better than reporting failure while leaving a wedged daemon running.

**Auth token lifecycle — fail-closed in the right direction.** `refreshToken` only maps `HttpError` into a categorized failure and rethrows anything else; `performRefresh` signs out solely on `kind === 'client'` (explicit rejection), clearing the in-memory ref, stored token, user, and step together, while network/server/parse failures preserve the deliberate offline-access path. Two accepted tradeoffs to be aware of: any 4xx (e.g., a proxy 403 that is not a real token rejection) de-authenticates — for auth, erring toward sign-out is correct; and a 200 body with `success: false` and no error remains a silent no-op, same as before this PR.

**`httpRpc` timeout — correct.** `AbortController` with `clearTimeout` in `finally`, abort mapped to an actionable message, 401/403 distinguished. One nit, not a blocker: the 30s budget covers only through response headers — a daemon that returns headers then stalls the body would still hang `res.json()` indefinitely. If it ever bites, pass the same signal's deadline through body consumption.

**Confirm dialogs and degraded-mode honesty — no concerns.** The mock deploy id sentinel matches the established mock-secret style, and `markDegraded` calls make browser-mode fabrication visible instead of silent.

**Tray icon fallback** — correct; startup panic replaced with a degraded-but-usable tray.

No secret material, no injection surface, no privilege change anywhere in the diff. The PR body's claims match the code line for line, including its own disclosure that the sensitivity self-check does not gate these paths.

Verdict: **approve** (`sec-review:approved`). Nits above are follow-up material, none blocking.

PR #261, opened 2026-07-10
fix(daemon): session.end lets any unsigned peer kill another agent's PTYs

RevealUIStudio:
Closes the HIGH cross-agent eviction in GAP-288 work-step 4. This is reachable **unsigned, by any socket peer**, today.

## The bug

`session.end` (`server.ts:1022`) resolved its target as `params.sessionId ?? params.agentId ?? ctx.agentId`, and was **not** in `MUTATING_OR_CONTENT_METHODS`. The dispatch identity gate is satisfied by any bare `actorAgentId` string, while the handler read `sessionId`. So one unsigned call:

```
session.end { actorAgentId: "anything", sessionId: "<victim>" }
```

reaches `notifyAgentEnded(victim)`, which evicts the victim's registered project roots (`filegit.ts` `evictRootsForAgent`) and kills every PTY it owns (`spawn.ts` `onAgentEnded`).

Only signed, client-owned identities can own roots and PTYs, so the destructive payload only ever lands on exactly the identities that were supposed to be protected.

## The fix

- `session.end` added to `MUTATING_OR_CONTENT_METHODS`, and to `signing.rs requires_signature()` in lockstep (both drift tests updated).
- The handler now **self-scopes to the verified signer** (`ctx.agentId`, guaranteed `boundVia === 'signature'` by the dispatch gate) and ignores caller-supplied targets entirely.
- The `params.sessionId ?? params.agentId` override is **deleted, not gated behind an admin role**. It had no caller: `harness.prune` reaps aged sessions by calling `notifyAgentEnded` directly (`server.ts:273`), and every real caller passes its own id. It existed only as the attack surface.

## Two things GAP-288 got wrong

The gap file's step 4 is right about the mechanism and wrong about the reasoning, so recording it here.

1. **"Hooks can't sign, so adding this to the signed set would stop sessions being ended."** `~/.claude/hooks/stop.js:107` calls `session.end` on a one-shot unsigned socket with no `actorAgentId`, so it already returns `-32002` and the hook swallows it. The hook's call has been a no-op for some time, and row cleanup was never its job (the GAP-153 periodic prune does that). No new breakage: `-32002` before, `-32003` after.
2. **"Or gate a cross-agent override behind an explicit admin role."** There is no admin caller. Building the role would be YAGNI on an attack surface.

The real reason the signed-set route is correct is a consumer the gap never mentions: **Studio**. It is client-owned and auto-signs exactly the methods in `requires_signature` (`harness.rs:158`). Adding `session.end` there makes Studio sign it with **zero Studio code change**.

## How I know that, rather than assume it

My first attempt self-scoped via `requireVerifiedAgent` *without* adding `session.end` to the signed set, on the theory that it would preserve the unsigned hook call. `pnpm --filter @revdev/daemon test` passed 541/541. Then `cargo test --test harness_integration` failed:

```
session.end succeeds: "client-owned identity studio-… requires a signed request for this method"
```

That approach would have broken Studio's live "end session" path, and the daemon suite could never have caught it. Reverted and rebuilt along the route above.

## Tests

Two new adversarial regressions in `adversarial-isolation.test.ts` §6.j:

- an unsigned peer cannot end another agent's session, **and the victim's root is still owned afterward** (proving `notifyAgentEnded` never ran)
- a *signed* agent naming a different target ends only its own session; the victim is untouched

**Both were verified to fail against the vulnerable handler** and pass against the fix. A regression test that cannot fail is not a regression test.

Two existing tests (`neon-sync`, `coordination` prune) called `session.end` in exactly the vulnerable shape (`actorAgentId` + `sessionId`, unsigned). They now sign. Notably `coordination.test.ts:925` already contained the comment *"the session.end cross-agent hole must NOT be recreated"* — the codebase knew.

`pnpm --filter @revdev/daemon test`: 544 passed, 27 files. `cargo test --lib signing` (drift test) green. `harness_integration`, `daemon_ctl_integration`, `ssh_output_pump` green. `pnpm lint` and `pnpm typecheck` clean.

## Deliberately not included

- **`session.update`'s cross-agent `task`/`files` write** (`server.ts:1048`, same pattern). It is knowingly deferred in a code comment, and `coordination.test.ts` explicitly asserts a signed caller *can* update a peer's task while `state` self-scopes. That is intentional behavior with a test behind it, so changing it is a decision, not a cleanup. I tried it, broke that test, and reverted. Flagged for GAP-288 rather than changed unilaterally.
- **`events.log` attribution spoof** (`server.ts:1492`) — log attribution only, no capability. Same reasoning: flagged, not folded in.
- The **command allow-list** (GAP-288 work-step 2) remains behind its owner design gate.

## Review

This is an exec/identity surface and I authored it. The guardrail-2 checker flags it (`signing.rs`). Not self-merging; it needs a recorded verdict from someone who is not me.

### DIFF
--- a/apps/studio/src-tauri/src/signing.rs
+++ b/apps/studio/src-tauri/src/signing.rs
@@ -520,6 +520,7 @@ mod tests {
         assert!(requires_signature("agent.input"));
         assert!(requires_signature("agent.resize"));
         assert!(requires_signature("agent.output"));
+        assert!(requires_signature("session.end"));
         assert!(!requires_signature("ping"));
         assert!(!requires_signature("session.list"));
     }
--- a/apps/studio/src-tauri/src/signing.rs
+++ b/apps/studio/src-tauri/src/signing.rs
@@ -84,6 +84,10 @@ pub fn requires_signature(method: &str) -> bool {
             | "agent.input"
             | "agent.resize"
             | "agent.output"
+            // session.end evicts the target's project roots and kills its PTYs.
+            // Signature-required so the daemon can self-scope it to the verified
+            // signer instead of a caller-supplied `sessionId`.
+            | "session.end"
     )
 }
 
--- a/packages/daemon/src/__tests__/adversarial-isolation.test.ts
+++ b/packages/daemon/src/__tests__/adversarial-isolation.test.ts
@@ -551,7 +551,12 @@ describe('adversarial agent isolation (B6 §6)', () => {
       await writeFile(anchor, `${anchorContent}${victimId}:${fpVictim}\n`);
 
       const rpcVictim = (method: string, params: Record<string, unknown>) =>
-        rpcFrame(socketPath, method, params, signFor(victimId, didVictim, fpVictim, kpVictim, method, params));
+        rpcFrame(
+          socketPath,
+          method,
+          params,
+          signFor(victimId, didVictim, fpVictim, kpVictim, method, params),
+        );
 
       try {
         await rpcFrame(socketPath, 'session.register', {
@@ -594,7 +599,12 @@ describe('adversarial agent isolation (B6 §6)', () => {
       await writeFile(anchor, `${anchorContent}${victimId}:${fpVictim}\n`);
 
       const rpcVictim = (method: string, params: Record<string, unknown>) =>
-        rpcFrame(socketPath, method, params, signFor(victimId, didVictim, fpVictim, kpVictim, method, params));
+        rpcFrame(
+          socketPath,
+          method,
+          params,
+          signFor(victimId, didVictim, fpVictim, kpVictim, method, params),
+        );
 
       try {
         await rpcFrame(socketPath, 'session.register', {
--- a/packages/daemon/src/__tests__/adversarial-isolation.test.ts
+++ b/packages/daemon/src/__tests__/adversarial-isolation.test.ts
@@ -536,6 +536,88 @@ describe('adversarial agent isolation (B6 §6)', () => {
         await rm(repo, { recursive: true, force: true });
       }
     });
+
+    it('an unsigned peer cannot end another agent’s session (no eviction, no kill)', async () => {
+      // The hole: session.end used to read `params.sessionId` verbatim while the
+      // identity gate was satisfied by any `actorAgentId` string. One unsigned
+      // call evicted the victim's roots and killed its PTYs.
+      const repo = await initRepo('iso-j-crossagent-');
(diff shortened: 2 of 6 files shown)


RevealUIStudio:
sec-review: approved — owner authorized the merge directly (2026-07-10), which is the recorded non-author verdict guardrail 2 requires. Noting it on the PR rather than only in chat, since that side-channel gap is exactly what the #1711 miss was made of.

RevealUIStudio:
Post-merge adversarial review by a non-author session (repairing the self-clear: this PR merged with `reviews=0`). Read against `packages/daemon/src/{server,filegit,eviction,spawn}.ts`, not the description.

**Verdict on this PR: CONFIRMED CORRECT.** The handler (`server.ts:1030`) self-scopes to `ctx.agentId` and requires `ctx.boundVia === 'signature'`; `session.end` is in `MUTATING_OR_CONTENT_METHODS` (`server.ts:219`) so the dispatch gate (`server.ts:1936`) rejects anything but a verified envelope before the handler runs; the caller-supplied `sessionId`/`agentId` override is deleted, not gated. Studio signs it in lockstep (`signing.rs`). The two regressions (unsigned attacker, signed cross-agent) both assert the victim keeps its root. Verifier binding checks method + paramsHash + ts-window + a replay nonce, and key enrollment is trust-anchored, so the signer identity cannot be spoofed. Good fix.

**But the remediation is incomplete, and this PR's own justification points at the gap.** The GAP-288 writeup deletes the caller-target on the grounds that `session.end` "had no caller — harness.prune calls notifyAgentEnded directly." `harness.prune` is itself `IDENTITY_EXEMPT` (`server.ts:117`), NOT in `MUTATING_OR_CONTENT_METHODS`, and takes a caller-controlled `staleDays` that `runPrune` only clamps with `Math.max(0, …)` (`server.ts:267`). So an unsigned `harness.prune {staleDays:0}` from any same-UID socket peer ends **every** live session and, via `notifyAgentEnded` → `evictRootsForAgent` + `pty.kill()` (`filegit.ts:89`, `spawn.ts:117`), evicts every agent's roots and kills every PTY. That is the blast radius this PR fixed for one victim, fleet-wide, still open.

Provenance: prune's exempt status predates this PR (GAP-153/#23), so this is not a regression here — it is the next door to the same primitive. Filed as **GAP-312** (revealui-jv#365) with a repro and a fix path (signature-gate the on-demand RPC; the periodic internal sweep calls `runPrune` directly and is unaffected).

PR #262, opened 2026-07-10
fix(daemon): signature-gate harness.prune + floor its thresholds (GAP-312)

RevealUIStudio:
Closes the fleet-wide eviction hole in `harness.prune` (GAP-312).

## The hole

`harness.prune` reaches the **same** eviction primitive as `session.end`, `notifyAgentEnded`, which fires `evictRootsForAgent` (`filegit.ts`) and the PTY-kill hook (`spawn.ts`). But it fans that across **every matched session**, not one. And it was:

- `IDENTITY_EXEMPT` (`server.ts`), so the dispatch identity gate returned early,
- absent from `MUTATING_OR_CONTENT_METHODS`, so the signature gate never fired, and
- schema-permissive: `staleDays` was `z.number().optional()`, the handler clamped only with `Math.max(0, staleDays)`.

So one unsigned frame from any same-UID socket peer (on a dev box: any dependency postinstall, any tool running as the operator):

```
{"jsonrpc":"2.0","id":1,"method":"harness.prune","params":{"staleDays":0}}
```

set the WHERE clause to `started_at < NOW()`, ended **every live session**, and killed **every PTY**. A negative `staleDays` clamped to 0 and did the same; `hardDeleteDays: 0` also hard-deleted the rows.

This is the exact blast radius revdev#261 closed for `session.end` (one victim), left open fleet-wide through the sibling door. The primitive, not the individual RPC, is the surface, which is why `session.end` alone was not enough.

## The fix

- **Signature-gate it.** Moved out of `IDENTITY_EXEMPT`, into `MUTATING_OR_CONTENT_METHODS`, mirrored in `signing.rs requires_signature()` (the cross-language contract the daemon set requires). Dispatch now rejects unsigned callers with `-32003` before the handler; the handler adds a `boundVia==='signature'` backstop, mirroring `spawn.ts requireAgent`.
- **Floor the thresholds at 1 day** in both the schema (`z.number().min(1)`) and `runPrune` (`Math.max(1, ...)`). A zero floor is not a defense, it is the fleet-wide selector. The `runPrune` floor also covers the env-var path (`REVDEV_STALE_THRESHOLD_DAYS` / `REVDEV_HARD_DELETE_DAYS`).
- The **periodic internal sweep is unaffected**: it calls `runPrune` directly (`server.ts:1718,1724`), never through the RPC.

## Tests, proven red first

Per `feedback-prove-tests-fail-without-the-fix`, the two new adversarial tests were shown **red against the pre-fix handler** before landing:

- an UNSIGNED `harness.prune` (valid-schema threshold) is rejected `-32003` and evicts nothing;
- a SIGNED `harness.prune` with `staleDays: 0` is rejected by the schema floor and evicts nothing.

The prior test `clamps negative thresholds to zero (defensive)` asserted the vulnerability as intended behavior (unsigned `staleDays: -100`, expected to age every session). It is removed. It was the same species as the mocked-`node:fs.stat` test revdev#260 deleted.

The three existing prune tests (age-out, hard-delete, health) now sign the call and age rows by **backdating** `started_at`/`ended_at` through the `_db` handle instead of sub-day thresholds + sleeps (faster, no flake). `validation.test.ts`, `signature-gate.test.ts`, and `signature-telemetry.test.ts` updated for the new behavior.

## Verification

- `@revdev/daemon`: **555/555** green.
- `cargo test --test harness_integration`: **4/4** green (serial; one parallel-only socket-timing flake on the unrelated `ping`/restart test, confirmed not caused by this change and passing in isolation and serially).
- `signing.rs`: `requires_signature_matches_daemon_set` asserts `harness.prune` signed and `harness.health` unsigned.

Studio never calls `harness.prune`, so adding it to `requires_signature()` is pure hardening with no client-path change.

## Review gate

This touches `signing.rs` and the daemon eviction surface, so it is **security-sensitive under guardrail 2** (`dup-work-and-review-gates.md`). It was authored by a Fable session; **it must not be self-cleared.** It needs a recorded verdict from a non-author reviewer (`sec-review:approved`, applied by the owner or a non-author coordinator) before merge. Merges to `test`, per branch flow.

Refs GAP-312. Sibling of GAP-288 (session.end, revdev#261).

### DIFF
--- a/apps/studio/src-tauri/src/signing.rs
+++ b/apps/studio/src-tauri/src/signing.rs
@@ -88,6 +88,13 @@ pub fn requires_signature(method: &str) -> bool {
             // Signature-required so the daemon can self-scope it to the verified
             // signer instead of a caller-supplied `sessionId`.
             | "session.end"
+            // harness.prune reaches the SAME eviction primitive as session.end,
+            // but fans it across every matched session rather than one. It was
+            // identity-exempt and unsigned, so one frame from any same-UID
+            // socket peer ended the whole fleet's sessions and killed every PTY
+            // (`staleDays: 0` selects `started_at < NOW()`). GAP-312.
+            // MUST mirror server.ts MUTATING_OR_CONTENT_METHODS.
+            | "harness.prune"
     )
 }
 
@@ -521,7 +528,12 @@ mod tests {
         assert!(requires_signature("agent.resize"));
         assert!(requires_signature("agent.output"));
         assert!(requires_signature("session.end"));
+        // GAP-312: same eviction primitive as session.end, fleet-wide fan-out.
+        assert!(requires_signature("harness.prune"));
         assert!(!requires_signature("ping"));
         assert!(!requires_signature("session.list"));
+        // harness.health is a read; it stays unsigned. Guards against a
+        // copy-paste that gates the wrong half of the harness.* surface.
+        assert!(!requires_signature("harness.health"));
     }
 }
--- a/packages/daemon/src/__tests__/coordination.test.ts
+++ b/packages/daemon/src/__tests__/coordination.test.ts
@@ -344,36 +344,43 @@ describe('GAP-153: stale-session prune', () => {
     expect(typeof health.prune?.lastDeletedCount).toBe('number');
   });
 
+  // harness.prune is signature-required (GAP-312). These tests sign the call,
+  // and because the threshold is now floored at 1 day they age rows by
+  // BACKDATING `started_at` / `ended_at` through the _db handle rather than by
+  // passing a sub-day threshold and sleeping. Backdating is both correct (it is
+  // what a genuinely stale session looks like) and faster (no real-time waits,
+  // no flake).
+
   it('ages out a session whose age exceeds the stale threshold', async () => {
-    // Register a stable-id session, then wait briefly so its started_at
-    // is older than the threshold we'll pass to prune.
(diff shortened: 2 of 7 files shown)


RevealUIStudio:
Non-author security verdict (I filed GAP-312; I did not write this fix): **sec-review:approved**. Adversarial read of the full diff.

The fix is correct and defense-in-depth. `harness.prune` is removed from `IDENTITY_EXEMPT` and added to `MUTATING_OR_CONTENT_METHODS`, so the dispatch gate rejects an unsigned caller with `-32003` before the handler runs; `signing.rs` mirrors it in `requires_signature()` (and adds a negative assertion that `harness.health` stays unsigned, guarding against over-gating). The `staleDays` amplifier is floored at 1 day in three places — schema, handler, and `runPrune` — and the `runPrune` floor also closes the env-var path (`REVDEV_STALE_THRESHOLD_DAYS`). The periodic internal sweep is untouched because it calls `runPrune` directly, never the RPC.

The tests carry the weight: an `unsigned harness.prune` is rejected and the victim session survives; a signed sub-day threshold is rejected by the floor; the signed-method exact-set includes it. Notably the PR rewrote a prior test that had *asserted the vulnerability as intended* (an unsigned `staleDays:-100` expecting success) — that is the prove-red discipline applied to the fix itself. CI green. Closes the fleet-wide eviction hole; merging.

PR #263, opened 2026-07-10
ci: add prove-red gate (verification-enforcement Phase 1a)

RevealUIStudio:
Phase 1a of the verification-enforcement lane. Makes "prove red" a machine-checked CI gate.

## What it does

A test that passes whether or not the fix is present proves nothing. On a PR that changes **both** product source and TS test files, the `prove-red` job reverts the source diff to the merge base, re-runs each changed test file, and requires **at least one to be red against the base**. A PR that changes source but carries no failing-first test is blocked.

## Design, and its honest limits

- **Partition** by path suffix, zero authored regex: TS test files vs code-source files. Fires only when both change (test-only, source-only, and docs PRs pass through).
- **Revert** code-source files to the merge base (added → `rm`, modified/deleted → `checkout base`), then run each changed test file per its owning package with vitest.
- **PASS iff ≥1 changed test file is red against base.** Green files are *reported* (compat update or inert) but do not block, because the gate cannot distinguish a legitimate compatibility update from an inert test at file granularity. Requiring *every* file to be red would falsely block PRs like GAP-312's, which legitimately updated an existing test for API compatibility.
- **Documented limits** (matched to the lane spec §4.2): file granularity, not per-test; a compile-failure-against-base counts as red, so it kills the worthless regression guard without certifying test *strength*; Rust/Go changed-test prove-red is **not** covered in Phase 1a and is logged loudly, never silently skipped.

## Validated both directions against the real GAP-312 branch (#262)

- **Positive:** 3 of 4 changed daemon test files red against base (they prove the fix); `signature-telemetry.test.ts` correctly reported as a compat update, not blocked; exit 0.
- **Negative:** a synthetic inert test (`expect(1 + 1).toBe(2)`, independent of its source change) is blocked, exit 1, "NONE of the changed test files fail against the base."

The prove-red tool was itself proven red and green, which is the discipline it enforces.

## Not in this PR (owner-gated)

Added as a **non-required** check. Making it a **required status check with admin bypass off** is the ruling that turns it from advisory to enforced (lane Phase 1c). This PR does not flip it, and the verification block (Phase 1b) is separate.

## Review

The fleet security checker reports this as not security-sensitive (CI + test tooling, no sensitive path). But it changes how the fleet verifies, so a non-author read before merge is worthwhile even though no gate requires it.

Lane + spec live in `revealui-jv` (`docs/lanes/verification-enforcement/plan.md`, `docs/specs/2026-07-10-verification-enforcement-prove-red-gate.md`).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -82,6 +82,36 @@ jobs:
       - run: pnpm install --frozen-lockfile
       - run: pnpm test
 
+  prove-red:
+    name: Prove-Red (changed tests must fail against base)
+    runs-on: ubuntu-latest
+    needs: quality
+    # PR-only: a merge-base is meaningless on a direct push to test/main.
+    if: github.event_name == 'pull_request'
+    steps:
+      # Check out the PR head (not the synthetic merge commit) with full history
+      # so scripts/ci/prove-red.mjs can compute the merge base and revert the
+      # source diff. See docs (verification-enforcement lane) for the mechanism.
+      - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
+        with:
+          ref: ${{ github.event.pull_request.head.sha }}
+          fetch-depth: 0
+      - uses: pnpm/action-setup@0ebf47130e4866e96fce0953f49152a61190b271 # v6.0.9
+      - uses: actions/setup-node@48b55a011bda9f5d6aeb4c2d9c7362e8dae4041e # v6
+        with:
+          node-version: 24
+          cache: pnpm
+      - run: pnpm install --frozen-lockfile
+      # Changed daemon tests import @revdev/protocol and @revdev/daemon; build
+      # them from source first, exactly as the root `test` script does, or the
+      # per-file vitest runs fail to resolve imports rather than on assertions.
+      - run: pnpm --filter @revdev/protocol build
+      - run: pnpm --filter @revdev/daemon build
+      - name: Prove changed tests are red against base
+        env:
+          BASE_REF: ${{ github.event.pull_request.base.sha }}
+        run: node scripts/ci/prove-red.mjs
+
   build:
     name: Build
     runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


RevealUIStudio:
Post-merge non-author review (I did not author this; it merged `reviews=0`, so this is the second-party read the gate itself should have had — same repair pattern as #260/#261).

**Verdict: SOUND as advisory (Phase 1a).** The prove-red semantics are correct: it fires only when a PR changes both product source and test files, reverts the source to the merge base, re-runs each changed test file, and blocks only when NONE of them is red against base — which is exactly "a passing test proves nothing until it fails against the old code," mechanized. It is honest about its limits in-code (file-granularity not per-test; a compile-fail counts as red so it does not certify test *strength*; green changed files are reported, not blocked). Zero authored regex. Nice correctness touch: it builds `@revdev/protocol` + `@revdev/daemon` before the per-file vitest runs, so an import failure cannot masquerade as a red assertion. It is a non-required job today, so it cannot deadlock anything yet.

**Three things to fix BEFORE this becomes a required check (the owner-gated flip):**

1. **The refactor exception is referenced in prose but not consumed by the gate.** `prove-red.mjs` names the `verify:no-behavior-change` label in its failure message but never reads it. A pure refactor has correctly-green tests, so once this check is *required* it will block every legitimate refactor. Before the flip, the gate must honor that label — and, per the verification-enforcement + review-eval convergence, the label must be applied by a NON-AUTHOR (the one-door, agent-can-never-self-apply shape). This is a hard sequencing dependency on the flip.
2. **The test/source classification is a trust boundary and is game-able** (GAP-313 territory). A "fix" placed in a `.test.ts` file is not reverted; product logic hidden in test files can dodge the gate, and a source change with no test file skips it entirely. `prove-red.mjs` is itself a gate-definition file — it needs CODEOWNERS / tamper-evidence per GAP-313 so it cannot be weakened in the same PR it gates.
3. **Rust/Go changed tests are surfaced-but-uncovered.** Today's revdev#261/#262 were exactly Rust+TS daemon *security* fixes, so the daemon's Rust surface being outside prove-red is a real coverage gap for the highest-value surface. Phase 1b.

None of these block the merge (it is advisory); they gate the required-check flip, which is the owner's call.

PR #82, opened 2026-05-19
chore(test): backflow main into test (1 commit behind — promotion-gate workflow merge)

RevealUIStudio:
Backflow main's promotion-gate workflow merge commit (e5c7b76b4) onto test.

Replaces closed PR #81 (which hit the head=main + SKIPPED required-checks gotcha per memory feedback_main_promotion_skipped_required_checks).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #270, opened 2026-07-11
fix(daemon): GAP-320 confinement Phase 1b — refuse secret-overlapping grants, deny REVDEV_SPAWN_CONFINEMENT passthrough

RevealUIStudio:
## What

Implements the approved GAP-320 Phase-1b confinement design ([.jv jv#411](https://github.com/RevealUIStudio/revealui-jv/pull/411), non-author verdict recorded, merged to `.jv:test`). Two hardening residuals surfaced by the revdev#264 crown-jewel adversarial review — neither a live vulnerability in the shipped design, both places the confinement layer could silently defeat itself under operator action.

### (a) Granted-root overlap guard
- `neverBoundSet(operatorHome)` materializes spec §4.3's never-bound list in code for the first time: the home-relative secrets (`.ssh`, `.age-identity`, `.revealui/passage-store`, `.config/gh`, `.npmrc`, `.aws`, `.docker/config.json`, `.claude`), the absolute NTFS mounts (`/mnt/c`, `/mnt/e`), and `/run/user/<uid>` (the ssh-agent socket). Each entry in **lexical + realpath-if-exists** form, because `repoReal` arrives realpath'd and a symlinked secret dir (`~/.ssh -> /data/ssh`) would slip a lexical-only compare. Computed per spawn, so a secret path appearing after daemon start is guarded without a restart.
- `assertGrantedRootBindable(repoReal, operatorHome)` refuses (throws a named error) when the granted root **is, contains, or lives inside** a secret path or the operator home. **Fail closed, not mask** — a mask that misses a path exposes it silently; an over-broad refusal fails loud and is fixed the same day (the design §3.1 records the full rationale).
- Runs at the top of `spawnConfined`, co-located with the tmpfs-then-bind sequence (confinement.ts §4.3) that creates the hazard. The `ConfinementBackend` contract now requires every backend to call it, so the staged darwin backend inherits the requirement.

### (b) `REVDEV_SPAWN_CONFINEMENT` env deny
`filterCallerEnv` rejects `REVDEV_SPAWN_CONFINEMENT` by name, checked **before** the `REVDEV_` prefix allowance, so a caller cannot seed the daemon's escape-hatch key into a sandboxed child's environment. Inert today (the resolver reads only the daemon's own `process.env`), but a caller-seedable boundary-control key in a sandboxed env invites a future misread. The `REVDEV_` prefix otherwise stands.

## Verification

- **Prove-red (per feedback-prove-tests-fail-without-the-fix):** stashed the source fix, ran the tests — all 14 new-behavior assertions fail against the pre-fix backend (refusals, `neverBoundSet`, the backend guard wiring, the env deny); they pass with the fix. Separator-safety pinned: `/base/op-scratch` and `.sshkeep` do **not** false-match; the normal `~/revfleet/<repo>` case is not refused.
- **Full daemon suite:** 612/612 green. **confinement.test.ts:** 47/47. `tsc --noEmit` clean. Biome clean.
- Zero authored regex — `join` + string equality + the `within()` prefix shape only. Fixtures use `/base/op`, never a literal `/home/<user>` (leak-scan-safe, per revdev#265).

## Scope / not in scope

Exactly the two GAP-320 items. The adjacent `file.*` reachability finding (a broad grant can read secrets at the file layer with no spawn) and the `project.open`-accepts-non-repos drift are **out of scope** and filed as [GAP-326](https://github.com/RevealUIStudio/revealui-jv/blob/test/docs/gaps/GAP-326.yml) (design §6). Shared-cache rw exposure is noted for Phase 2.

## Review ask (crown-jewel — do NOT self-merge)

`agent.spawn` is a crown-jewel surface. This PR carries **no gate labels**. Requesting a non-author guardrail-2 verdict grounded to file:line before merge. No live Fable session (safeguard flip 2026-07-11); per model-allocation the review falls back to an Opus peer, recorded. Owner merges on the recorded verdict.

Design: `.jv docs/gap-specs/GAP-320-confinement-phase-1b-hardening-design.md`

### DIFF
--- a/packages/daemon/src/__tests__/confinement-integration.test.ts
+++ b/packages/daemon/src/__tests__/confinement-integration.test.ts
@@ -105,6 +105,7 @@ describe.skipIf(!RUN)('confinement integration (real bwrap)', () => {
       cwd: repoReal,
       agentHome,
       operatorHome,
+      dataDir: join(operatorHome, 'daemon-data'),
     });
     const r = spawnSync(file, argv, { encoding: 'utf8', timeout: 15_000 });
     return { status: r.status, out: `${r.stdout ?? ''}${r.stderr ?? ''}` };
@@ -160,12 +161,14 @@ describe.skipIf(!RUN)('confinement integration (real bwrap)', () => {
     // name before any sandbox is built — while the sibling happy-path spawn above
     // still succeeds (DAILY DRIVER). The guard fires in spawnConfined, so this
     // exercises the real entrypoint with a real secret-bearing home on disk.
+    const dataDir = join(operatorHome, 'daemon-data');
     expect(() =>
       backend.spawnConfined('/bin/sh', ['-c', 'true'], {
         repoReal: operatorHome,
         cwd: operatorHome,
         agentHome,
         operatorHome,
+        dataDir,
       }),
     ).toThrow(/is or contains the operator home/);
     expect(() =>
@@ -174,6 +177,7 @@ describe.skipIf(!RUN)('confinement integration (real bwrap)', () => {
         cwd: join(operatorHome, '.ssh'),
         agentHome,
         operatorHome,
+        dataDir,
       }),
     ).toThrow(/overlaps the never-bound secret path/);
   });
--- a/packages/daemon/src/__tests__/confinement-integration.test.ts
+++ b/packages/daemon/src/__tests__/confinement-integration.test.ts
@@ -154,6 +154,30 @@ describe.skipIf(!RUN)('confinement integration (real bwrap)', () => {
     expect(out).toContain('REPO-VISIBLE');
   });
 
+  it('refuses a spawn whose granted root overlaps a real fixture secret (GAP-320a §5.D)', () => {
+    // End-to-end against the seeded operator-home layout: a granted root that IS
+    // the operator home, or that lives inside the seeded ~/.ssh, is refused by
+    // name before any sandbox is built — while the sibling happy-path spawn above
+    // still succeeds (DAILY DRIVER). The guard fires in spawnConfined, so this
+    // exercises the real entrypoint with a real secret-bearing home on disk.
+    expect(() =>
+      backend.spawnConfined('/bin/sh', ['-c', 'true'], {
+        repoReal: operatorHome,
+        cwd: operatorHome,
+        agentHome,
+        operatorHome,
+      }),
+    ).toThrow(/is or contains the operator home/);
+    expect(() =>
+      backend.spawnConfined('/bin/sh', ['-c', 'true'], {
+        repoReal: join(operatorHome, '.ssh'),
+        cwd: join(operatorHome, '.ssh'),
+        agentHome,
+        operatorHome,
+      }),
+    ).toThrow(/overlaps the never-bound secret path/);
+  });
+
   it('HOME points at the agent home, not the operator home', () => {
     const { out } = confined('echo "HOME=$HOME"');
     expect(out).toContain(`HOME=${agentHome}`);
--- a/packages/daemon/src/__tests__/confinement.test.ts
+++ b/packages/daemon/src/__tests__/confinement.test.ts
@@ -115,6 +115,7 @@ describe('linuxBubblewrapBackend.spawnConfined', () => {
     cwd: '/base/op/repo/packages/x',
     agentHome: '/base/op/.local/share/revealui/agent-homes/deadbeef',
     operatorHome: '/base/op',
+    dataDir: '/base/op/.local/share/revealui',
   };
 
   it('execs bwrap, not the raw command', () => {
@@ -297,33 +298,34 @@ describe('ensureAgentHome', () => {
 
 describe('assertGrantedRootBindable', () => {
   const HOME = '/base/op';
+  const DATADIR = '/base/op/.local/share/revealui';
 
   it('refuses the operator home itself', () => {
-    expect(() => assertGrantedRootBindable(HOME, HOME)).toThrow(/is or contains the operator home/);
+    expect(() => assertGrantedRootBindable(HOME, HOME, DATADIR)).toThrow(/is or contains the operator home/);
   });
 
   it('refuses an ANCESTOR of the operator home', () => {
     // repoReal = /base contains /base/op — the tmpfs-then-bind would carry the home.
-    expect(() => assertGrantedRootBindable('/base', HOME)).toThrow(
+    expect(() => assertGrantedRootBindable('/base', HOME, DATADIR)).toThrow(
       /is or contains the operator home/,
     );
   });
 
   it('refuses a granted root that IS a secret path', () => {
-    expect(() => assertGrantedRootBindable(join(HOME, '.ssh'), HOME)).toThrow(
+    expect(() => assertGrantedRootBindable(join(HOME, '.ssh'), HOME, DATADIR)).toThrow(
       /overlaps the never-bound secret path/,
     );
   });
 
   it('refuses a granted root INSIDE a secret path', () => {
-    expect(() => assertGrantedRootBindable(join(HOME, '.ssh', 'sub'), HOME)).toThrow(
+    expect(() => assertGrantedRootBindable(join(HOME, '.ssh', 'sub'), HOME, DATADIR)).toThrow(
       /overlaps the never-bound secret path/,
     );
   });
 
   it('refuses a granted root that CONTAINS a secret path (.revealui over passage-store)', () => {
     // repoReal = <home>/.revealui contains <home>/.revealui/passage-store.
-    expect(() => assertGrantedRootBindable(join(HOME, '.revealui'), HOME)).toThrow(
+    expect(() => assertGrantedRootBindable(join(HOME, '.revealui'), HOME, DATADIR)).toThrow(
       /overlaps the never-bound secret path/,
     );
   });
@@ -333,27 +335,51 @@ describe('assertGrantedRootBindable', () => {
     // itself (equal direction). Fixtures avoid the Windows-user and trailing-slash
     // mount shapes the private-leak scanner flags — the refusal is identical for
     // any path on those mounts.
-    expect(() => assertGrantedRootBindable('/mnt/c/project', HOME)).toThrow(
+    expect(() => assertGrantedRootBindable('/mnt/c/project', HOME, DATADIR)).toThrow(
       /overlaps the never-bound secret path/,
     );
-    expect(() => assertGrantedRootBindable('/mnt/e', HOME)).toThrow(
+    expect(() => assertGrantedRootBindable('/mnt/e', HOME, DATADIR)).toThrow(
+      /overlaps the never-bound secret path/,
+    );
+  });
+
+  it('refuses a granted root that IS the daemon data dir', () => {
+    expect(() => assertGrantedRootBindable(DATADIR, HOME, DATADIR)).toThrow(
+      /overlaps the never-bound secret path/,
+    );
+  });
+
+  it('refuses a granted root that CONTAINS the daemon data dir', () => {
+    // repoReal = <home>/.local/share contains the daemon dir under it.
+    expect(() => assertGrantedRootBindable(join(HOME, '.local', 'share'), HOME, DATADIR)).toThrow(
+      /overlaps the never-bound secret path/,
+    );
+  });
+
+  it('refuses the daemon data dir even when it is configured OUTSIDE the home', () => {
+    const outHome = '/var/lib/revealui-daemon';
+    expect(() => assertGrantedRootBindable(outHome, HOME, outHome)).toThrow(
+      /overlaps the never-bound secret path/,
+    );
+    // A root beneath it (e.g. the grants DB dir) is refused too.
+    expect(() => assertGrantedRootBindable(`${outHome}/pglite`, HOME, outHome)).toThrow(
       /overlaps the never-bound secret path/,
     );
   });
 
   it('does NOT refuse the normal case: a project root beneath the home', () => {
     // The supported shape — ~/revfleet/<repo>. Must NOT throw.
-    expect(() => assertGrantedRootBindable(join(HOME, 'revfleet', 'revealui'), HOME)).not.toThrow();
+    expect(() => assertGrantedRootBindable(join(HOME, 'revfleet', 'revealui'), HOME, DATADIR)).not.toThrow();
   });
 
   it('does NOT refuse a sibling of the home that shares a name prefix (separator-safe)', () => {
     // /base/op-scratch must not match /base/op via a naive startsWith.
-    expect(() => assertGrantedRootBindable('/base/op-scratch/repo', HOME)).not.toThrow();
+    expect(() => assertGrantedRootBindable('/base/op-scratch/repo', HOME, DATADIR)).not.toThrow();
   });
 
   it('does NOT refuse a repo whose name merely prefixes a secret name (.sshkeep)', () => {
     // <home>/.sshkeep is not <home>/.ssh — separator-safe within() must not match.
-    expect(() => assertGrantedRootBindable(join(HOME, '.sshkeep'), HOME)).not.toThrow();
+    expect(() => assertGrantedRootBindable(join(HOME, '.sshkeep'), HOME, DATADIR)).not.toThrow();
   });
 
   it('refuses a SYMLINKED secret dir via the realpath-if-exists form', async () => {
@@ -367,7 +393,7 @@ describe('assertGrantedRootBindable', () => {
       await symlink(target, join(home, '.ssh'));
       // realpath the target the same way requireRootAndDir would hand it in.
       const repoReal = await realpath(target);
-      expect(() => assertGrantedRootBindable(repoReal, home)).toThrow(
+      expect(() => assertGrantedRootBindable(repoReal, home, join(home, 'daemon-data'))).toThrow(
         /overlaps the never-bound secret path/,
       );
     } finally {
@@ -399,7 +425,7 @@ describe('resolveOperatorHome (GAP-320a review S1)', () => {
 
 describe('neverBoundSet', () => {
   it('materializes the home-relative secrets and absolute mounts', () => {
-    const set = neverBoundSet('/base/op');
+    const set = neverBoundSet('/base/op', '/base/op/.local/share/revealui');
     for (const p of [
       '/base/op/.ssh',
       '/base/op/.age-identity',
@@ -418,7 +444,7 @@ describe('neverBoundSet', () => {
 
   it('includes /run/user/<uid> when getuid is available', () => {
     if (typeof process.getuid !== 'function') return;
-    const set = neverBoundSet('/base/op');
+    const set = neverBoundSet('/base/op', '/base/op/.local/share/revealui');
     expect(set).toContain(`/run/user/${process.getuid()}`);
   });
 });
@@ -433,6 +459,7 @@ describe('linuxBubblewrapBackend.spawnConfined — overlap guard', () => {
     cwd: '/base/op/.ssh',
     agentHome: '/base/op/.local/share/revealui/agent-homes/deadbeef',
     operatorHome: '/base/op',
+    dataDir: '/base/op/.local/share/revealui',
   };
 
   it('refuses to build argv when the granted root overlaps a secret path', () => {
@@ -441,12 +468,23 @@ describe('linuxBubblewrapBackend.spawnConfined — overlap guard', () => {
     );
   });
 
+  it('refuses to build argv when the granted root overlaps the daemon data dir', () => {
+    expect(() =>
+      backend.spawnConfined('bash', [], {
+        ...base,
+        repoReal: '/base/op/.local/share/revealui',
+        cwd: '/base/op/.local/share/revealui',
+      }),
+    ).toThrow(/overlaps the never-bound secret path/);
+  });
+
   it('still builds argv for a normal project root beneath the home', () => {
     const { argv } = backend.spawnConfined('bash', [], {
       repoReal: '/base/op/revfleet/repo',
       cwd: '/base/op/revfleet/repo',
       agentHome: base.agentHome,
       operatorHome: base.operatorHome,
+      dataDir: base.dataDir,
     });
     expect(argv).toContain('--bind');
     expect(argv.join(' ')).toContain('--bind /base/op/revfleet/repo /base/op/revfleet/repo');
--- a/packages/daemon/src/__tests__/confinement.test.ts
+++ b/packages/daemon/src/__tests__/confinement.test.ts
@@ -301,7 +301,9 @@ describe('assertGrantedRootBindable', () => {
   const DATADIR = '/base/op/.local/share/revealui';
 
   it('refuses the operator home itself', () => {
-    expect(() => assertGrantedRootBindable(HOME, HOME, DATADIR)).toThrow(/is or contains the operator home/);
+    expect(() => assertGrantedRootBindable(HOME, HOME, DATADIR)).toThrow(
+      /is or contains the operator home/,
+    );
   });
 
   it('refuses an ANCESTOR of the operator home', () => {
@@ -369,7 +371,9 @@ describe('assertGrantedRootBindable', () => {
 
   it('does NOT refuse the normal case: a project root beneath the home', () => {
     // The supported shape — ~/revfleet/<repo>. Must NOT throw.
-    expect(() => assertGrantedRootBindable(join(HOME, 'revfleet', 'revealui'), HOME, DATADIR)).not.toThrow();
+    expect(() =>
+      assertGrantedRootBindable(join(HOME, 'revfleet', 'revealui'), HOME, DATADIR),
+    ).not.toThrow();
   });
 
   it('does NOT refuse a sibling of the home that shares a name prefix (separator-safe)', () => {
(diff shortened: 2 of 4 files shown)


RevealUIStudio:
## Guardrail-2 verdict: **APPROVE**

**Reviewer:** non-author, independent Opus reviewer. I did not write this code. This is a recorded adversarial guardrail-2 verdict on a crown-jewel surface (`agent.spawn` kernel boundary). The owner disposes (merge) on this verdict; I do not merge, label, or push.

**Grounded against** the PR branch @ `16bf6d3`, read and run in a checked-out worktree.

---

### What I verified (personally, not from the description)

**1. The guard is real and wired on the actual spawn path — not just defined.**
`agent.spawn` → `conf.backend.spawnConfined(...)` at `spawn.ts:292` → `linuxBubblewrapBackend.spawnConfined` calls `assertGrantedRootBindable(repoReal, operatorHome)` at `confinement.ts:330`, **before any argv is assembled**. The guard rides the one real call site.

**2. `repoReal` arrives fully normalized, so the obvious bypasses are closed upstream.**
`requireRootAndDir` → `requireRoot` does `realpath(expandTilde(repoPathRaw))` at `filegit.ts:219`. So `..` segments, trailing slashes, and a granted root that is itself a symlink-to-secret are all resolved to a canonical path before the guard sees it. The never-bound set additionally carries a **realpath-if-exists** form (`confinement.ts:260-266`), which I confirmed catches a symlinked secret dir (`~/.ssh -> <tmp>/real-ssh`) — the unit test at `confinement.test.ts:358-375` grants the symlink *target* and is refused. Verified.

**3. `within()` is separator-safe in both directions.**
`confinement.ts:37-39`: `target === root || target.startsWith(root + sep)`. `/base/op-scratch` does **not** match `/base/op`; `<home>/.sshkeep` does **not** match `<home>/.ssh`; `<home>/.ssh/sub` **does** match. All three tested (`confinement.test.ts:348-356, 317-321`).

**4. `cwd` and `agentHome` cannot be steered onto a secret independently of the guard.**
`cwd` is realpath'd and asserted `within(repoReal, real)` at `filegit.ts:317/267`, and the sandbox only `--chdir`s into it after binding `repoReal` — so `cwd` is always at/beneath an already-guarded root. `agentHome` is `<dataDir>/agent-homes/<sha256(agentId)>` (`confinement.ts:493-495`), daemon-controlled, hashed, no traversal. Neither is a hole.

**5. Env deny (b) runs before the prefix allowance and can't be ordered around.**
`filterCallerEnv` checks `CALLER_ENV_DENY_EXACT` **first** in the loop body (`confinement.ts:154`), before the `REVDEV_` prefix allow at `164-169`, and throws — so a benign `REVDEV_` key preceding it does not wave it through (tested `confinement.test.ts:52`). It is also defense-in-depth: the resolver reads only the daemon's own `process.env` (`spawn.ts:284`, no opts), never caller env, so the key was already inert in a child. Correct.

**6. Escape-hatch path correctly gets no guard.**
`mode:'none'` sets `file=command` directly (`spawn.ts:299-305`) and never reaches the backend, so the guard is structurally unreachable there — which is the intended semantics (whole FS already exposed + audited as `confinement='none'`). Fail-closed on no-bwrap / unimplemented platforms is unchanged (`confinement.ts:454-471`).

**7. Scope split is honest.** The confinement guard guards the *bind*, not the *registration*; `project.open` accepting a broad root so `file.read` reaches secrets with no spawn is explicitly routed out to the Part-B authorization surface (design §6.2), not silently dodged. Agreed.

### Prove-red (I ran it myself)
Reverted `confinement.ts` to its pre-fix parent (`HEAD~1`) while keeping the new test file, then ran the confinement suite:

- **Pre-fix: 14 failed / 33 passed** — every new-behavior assertion red (`assertGrantedRootBindable`/`neverBoundSet` `is not a function`, the overlap-guard `spawnConfined` refusal, and the `REVDEV_SPAWN_CONFINEMENT` env-deny).
- **With the fix restored: 47 / 47 passed.**
- **Full daemon suite: 612 / 612 passed** (29 files, ~65s).
- Worktree restored clean (`git status` empty).

Real-bwrap integration suite (`confinement-integration.test.ts`) still proves the boundary holds end-to-end (secret reads denied in-sandbox, repo readable, symlink-escape denied) and is honest about its prove-red controls.

---

### Non-blocking findings (hardening + completeness — none is a live vulnerability)

**S1 — `operatorHome` is not realpath'd before the guard (`spawn.ts:285`).**
`operatorHome = process.env.HOME ?? '/root'` flows un-normalized into `assertGrantedRootBindable`, which compares it against a **realpath'd** `repoReal`. The home-identity and ancestor checks (`confinement.ts:290`) are therefore lexical-only: if the daemon's `$HOME` ever contained a symlink component, `realpath(home)` would diverge from the lexical `operatorHome` and a grant of the realpath'd home would slip past the `===`/`within` home checks. It is **not exploitable on the fleet host** (HOME is a real directory, and the realpath-if-exists never-bound entries backstop it via any existing home-relative secret — `.ssh`/`.claude`/etc. all exist), so I am not blocking. But the design's own rationale (§3.2) is about matching normalization domains for exactly the symlink case, and the fix is one line: realpath `operatorHome` once before the guard and `neverBoundSet`, mirroring `repoReal`. Recommend closing it.

**S2 — design §5.D's real-bwrap overlap-refusal integration test was not delivered.** `confinement-integration.test.ts` covers the Phase-1 boundary and the happy path, but not the promised end-to-end "granted root overlapping a fixture secret is refused." Low value (the refusal is a pre-bwrap pure predicate, already exercised at the real `spawnConfined` entrypoint by the unit tests), but the §7 acceptance line "real-bwrap integration coverage → §5.D" is slightly overstated — either add the case or reconcile the design claim.

**Nit** — design §5.C test 11 (escape-hatch + overlapping root still spawns, pinning that the guard wasn't hoisted into the handler) isn't present. Structurally guaranteed, so it's only a missing regression-pin.

---

**Bottom line:** the security property is correct, wired on the real spawn path, prove-red verified (14 red → 0 red, 612/612 full suite), and I found no bypass in the covered surface. Fail-closed and refuse-not-mask posture is sound. S1/S2 are hardening/completeness, not live holes. **APPROVE.** Owner disposes — I do not merge.

RevealUIStudio:
## Author update — applied the review's S1 + S2 findings (commit ee093cb)

The independent non-author Opus review above returned APPROVE with two non-blocking recommendations; both are now addressed so the PR merges hardened:

- **S1 (realpath operatorHome):** `operatorHome` flowed un-realpath'd into a guard that compares against a realpath'd `repoReal`, making the home-identity/ancestor checks lexical-only. New `resolveOperatorHome()` realpaths it (lexical fallback when it doesn't resolve); `spawn.ts` uses it so the guard and the tmpfs bind agree on the canonical home. Unit-tested (symlinked home resolves; missing path falls back).
- **S2 (§5.D integration test):** delivered the promised real-fixture overlap-refusal case in `confinement-integration.test.ts` — against the seeded operator-home layout, a root that IS the home or lives inside the seeded `~/.ssh` is refused at the real `spawnConfined` entrypoint, while the happy-path daily-driver spawn still succeeds.

Re-verified: full-diff prove-red vs `origin/test` = **17 new assertions fail against base, all pass with the fix**; full daemon suite **615/615**; typecheck + biome + private-leak scan clean.

A second, genuinely independent live Opus session is also reviewing and will post its own verdict. Owner merges on the recorded verdict(s) — I have not merged or labeled.

RevealUIStudio:
## Independent non-author verdict — APPROVE the implementation, with ONE required addition before merge (Opus 4.8, live session)

Second independent adversarial pass (the first was a subagent Opus APPROVE whose S1/S2 are applied). Grounded in the branch at `ee093cb`, re-derived from the code rather than the prior review — I did NOT re-run the author's assertions, I re-checked the security properties myself.

### Verified correct
- **The guard is on the live spawn path, fail-safe placed.** `agent.spawn` → `conf.backend.spawnConfined` (spawn.ts:292) → `assertGrantedRootBindable(repoReal, operatorHome)` at the top of `spawnConfined`, **before any argv is assembled** (confinement.ts). The escape-hatch path (`conf.escapeHatch`, spawn.ts:299-305) never calls `spawnConfined`, so it is correctly unguarded — confinement is not authorization (spec §4.2 inv 3).
- **Predicate matches the approved design.** Refuses `repoReal === operatorHome` and `within(repoReal, operatorHome)` (home or ancestor); for each never-bound `nb`, refuses `within(repoReal, nb)` (secret inside the grant) or `within(nb, repoReal)` (grant is/inside a secret). The supported `within(operatorHome, repoReal)` beneath-home direction is correctly NOT refused — proven by the `does NOT refuse the normal case` test.
- **`within()` is a separate, separator-safe copy** kept local so confinement carries no filegit dependency; the `.sshkeep` and sibling-name-prefix tests prove it is not a naive `startsWith`.
- **S1 (resolveOperatorHome) is wired end to end.** spawn.ts:285 now passes `resolveOperatorHome()` (realpath'd) into the guard, so the home-identity/ancestor checks compare canonical paths against the already-realpath'd `repoReal`; the lexical fallback never widens the guard (never-bound entries carry their own realpath-if-exists forms). Correct.
- **Item (b) is exactly right:** `CALLER_ENV_DENY_EXACT` checked FIRST in `filterCallerEnv`, before the `REVDEV_` prefix allowance, throwing by name. The membership criterion (daemon-control keys only) is documented against future drift.
- **Tests are thorough and prove-red-shaped**, including the symlinked-secret realpath-if-exists case and the §5.D real-bwrap integration refusal (S2). Prove-red counts I take on the author's + first reviewer's record (17 assertions red at base); the structure supports it.

### One required addition before merge — the never-bound set omits the daemon data directory
This is the finding from the jv#411 design verdict that did **not** make it into the build. `neverBoundSet(operatorHome)` materializes spec §4.3's list (`.ssh`, `.age-identity`, `.revealui/passage-store`, `.config/gh`, `.npmrc`, `.aws`, `.docker/config.json`, `.claude`, `/mnt/c`, `/mnt/e`, `/run/user/<uid>`) but **not** `getDaemonConfig().dataDir` (default `~/.local/share/revealui`). That directory holds the daemon's **PGlite integrity DB** (`agent_identity`, `grants`, `project_roots` — `new PGlite(dataDir)`, cli.ts:112 / server.ts:1731), the **control socket**, and the **per-agent homes** — your own test fixture shows it: `agentHome: '/base/op/.local/share/revealui/agent-homes/deadbeef'` (confinement.test.ts:116).

Reachability, grounded in this branch: `project.open('~/.local/share')` (or the dataDir itself) is registrable — it is a non-repo, and `project.open` never rejects non-repos (filegit.ts:509-514, the §6.1 finding). `assertGrantedRootBindable('~/.local/share', operatorHome)` does **not** throw: it is beneath the home (the allowed direction), and no never-bound entry matches it. So an operator who grants that root and spawns re-binds the integrity DB + control socket **read-write inside the sandbox**, letting a confined agent rewrite its own `grants`/`agent_identity` rows (privilege escalation inside the daemon's own authorization model) or reach the control socket. This is the **same class and severity** as the `~/.ssh` overlap the guard already refuses — the guard's own cache-bind code even treats `~/.local/share` as secret-adjacent (confinement.ts:230-231), so the omission is an internal asymmetry, not a carve-out.

**Fix (clean, keeps confinement dependency-free):** thread `getDaemonConfig().dataDir` into the guard via `ConfinementOpts` the same way `operatorHome` is passed (spawn.ts already has `getDaemonConfig().dataDir` at :279), then add it to `neverBoundSet` in both lexical and realpath-if-exists forms. Sourcing from config covers a custom `REVDEV_DAEMON_DATA`; the default is the home-relative `.local/share/revealui`. No conflict with the separate `agentHome` bind — the guard inspects only `repoReal`, and `agentHome` is never a *granted root*. Add red tests: `repoReal = <dataDir>` and `repoReal = <home>/.local/share` are refused; keep the beneath-home non-regression. Update spec §4.3's source list in lockstep so the two don't drift again.

### Non-blocking note
The never-bound set is a spec-defined list; a broader credential-file review (`~/.netrc`, `~/.gnupg`, `~/.kube`) is a reasonable spec follow-up but is out of scope for this PR — don't expand it here beyond the dataDir, which is load-bearing for the daemon's own integrity.

### Verdict
The design realization is correct, well-wired, and well-tested, and S1/S2 are sound. **APPROVE conditional on one change: add the daemon dataDir to the never-bound set + a red test before merge.** A crown-jewel confinement guard shipping with a known hole in its never-bound set is exactly what the two-reviewer gate exists to catch, so I'm holding on that one item rather than clean-approving. With it applied, this is a clean crown-jewel merge on the recorded verdicts.

RevealUIStudio:
## Updated verdict (Opus 4.8, live session) — APPROVE. Required finding resolved.

Re-reviewed at `d79d201`. My one required change is fully addressed, cleanly:

- **`dataDir` threaded via `ConfinementOpts`** (confinement.ts:62) — keeps the confinement layer config-dependency-free, exactly the shape I recommended; `spawn.ts` sources it from `getDaemonConfig().dataDir` and passes it into `spawnConfined`.
- **Added to `neverBoundSet(operatorHome, dataDir)`** in the lexical array, so it flows through the realpath-if-exists loop → carried in **both** forms. Because it is a standalone entry (not home-relative), a **custom `REVDEV_DAEMON_DATA` outside the home is covered** — and there is a test for exactly that.
- **`assertGrantedRootBindable(repoReal, operatorHome, dataDir)`** signature + call-site updated; guard still runs before any argv.
- **Red tests are comprehensive:** refuses `repoReal = <dataDir>`, refuses `repoReal = <home>/.local/share` (contains it), refuses a dataDir configured outside the home, and keeps the beneath-home / `/base/op-scratch` / `.sshkeep` non-regressions. The `ConfinementOpts` doc records the key nuance — the per-agent home subdir under `dataDir` is still bound (it is the agent's own home); only a *granted root* overlapping `dataDir` is refused.

The daemon's own authorization state (grants, identities, roots), its control socket, and the per-agent homes can no longer be re-exposed read-write through a broad grant. **Verdict: APPROVE.** Both independent Opus passes now clear the confinement substance.

### One CI note (not a defect in this PR)
The `Quality` check is red, but on a **pre-existing, unrelated** biome lint: `lint/complexity/useLiteralKeys` at `apps/studio/src/hooks/use-sync.ts:82` (`state.errors['__all__']`). That file is **not in this PR's diff** and is byte-identical on `test`, so this PR did not introduce it — the branch simply inherits a lint violation already sitting on `test`. It needs a separate one-line fix (`biome check --write` on that file, or a literal `.__all__` access) so the branch CI goes green before merge; it has no bearing on the confinement change or this verdict.

RevealUIStudio:
## Author update — dataDir hole fixed (commit d79d201) + Quality-gate red is unrelated

**The independent live-session review's finding is fixed.** `neverBoundSet` now includes the daemon data dir (`getDaemonConfig().dataDir`, default `~/.local/share/revealui`), threaded through `ConfinementOpts` into the guard in both lexical and realpath-if-exists forms. A granted root that IS, CONTAINS, or lives inside dataDir is refused — including when dataDir is configured outside the operator home. Proven red-first (4 new dataDir assertions fail against base, pass with the fix); full daemon suite **619/619**; typecheck + biome + private-leak scan clean. Spec §4.3 + the GAP-320 design are updated in lockstep at **[jv#422](https://github.com/RevealUIStudio/revealui-jv/pull/422)**.

**The one red check (`Quality`) is NOT from this PR.** `origin/test` is already red on `biome check .`: biome 2.5.2 flags `state.errors['__all__']` in `apps/studio/src/hooks/use-sync.ts:82` (`useLiteralKeys`), a file this PR does not touch. It fails the gate on every PR against the repo. Split out and fixed in **[revdev#271](https://github.com/RevealUIStudio/revdev/pull/271)** (one line, dot access). Merge order: land #271 first, then this PR needs a branch update (merge `test` in) to inherit the fix and turn Quality green — a re-run alone won't clear it (base-side fix → fresh merge commit required).

Still no self-merge / no labels — owner disposes on the recorded verdicts.

RevealUIStudio:
## Guardrail-2 verdict: APPROVE (non-author, Fable 5)

Independent adversarial review of the GAP-320 Phase-1b build against `docs/gap-specs/GAP-320-confinement-phase-1b-hardening-design.md`, grounded to the PR head (`426482d`) and re-derived against `origin/test`. Not the author. Ran on Fable with no safeguard pause on this security-review content.

### What I verified against the spec

- **Refuse-over-mask, both overlap directions (§3.1, §3.3).** `assertGrantedRootBindable` fails closed and checks both `within(repoReal, nb)` (secret inside the granted root) and `within(nb, repoReal)` (granted root is or lives inside a secret), plus the home identity/ancestor case. `within()` is the separator-safe prefix shape (no bare `startsWith`), so `/base/op-scratch` and `~/.sshkeep` correctly do NOT match. The asymmetry is right: a root *beneath* the home is not refused.
- **Guard sits inside `linuxBubblewrapBackend.spawnConfined`, before any argv is assembled** (confinement.ts:368, ahead of the `--tmpfs`/`--bind` sequence it protects). One call site; the interface contract (lines 74-80) documents the requirement so the staged darwin backend inherits it explicitly.
- **Escape hatch deliberately unguarded** — confirmed *by construction*: `resolveConfinementBackend` returns `backend: null, escapeHatch: true`, and the escape-hatch branch in `agent.spawn` sets `file = command; argv = args` without ever calling `spawnConfined`. The guard cannot fire there, which is the intended §3.4 semantics (confinement is not authorization).
- **`REVDEV_SPAWN_CONFINEMENT` rejected by name in `filterCallerEnv`** (confinement.ts:163), checked FIRST — before the `REVDEV_` prefix allowance that would otherwise wave it through. Throws naming the key; the prefix allowance otherwise stands. Membership criterion (daemon-control keys only) is documented and correctly excludes the prune knobs.
- **Never-bound materialization matches spec §4.3 including the jv#422 dataDir addition.** `neverBoundSet` builds the eight home-relative secrets + `/mnt/c` + `/mnt/e` + `/run/user/<uid>` (conditional on `getuid`) + `dataDir`, each in lexical and realpath-if-exists form. `dataDir` gets its own entry and is refused even when configured outside the home (test covers `/var/lib/revealui-daemon`). jv#422 (MERGED) carries the matching spec text in both the gap-spec and the parent spec.
- **Symlinked-secret coverage (realpath-if-exists) and the S1 `resolveOperatorHome` realpath** are both present and tested with real on-disk symlink fixtures.

### Red-first proof (the #82/#403 standard — I re-ran it myself)

At PR head: `confinement.test.ts` 53/53, full daemon suite 619/619, `tsc --noEmit` clean.

Against a neutered tree (proving the tests actually bite):
- Removing the `assertGrantedRootBindable` call from `spawnConfined` → the 2 backend-wiring tests go RED.
- Stubbing the predicate body (home check off, never-bound loop empty) + disabling the env-deny branch → **13 acceptance tests go RED** (all overlap refusals across secret paths, ancestor, NTFS mounts, dataDir in/out of home, symlinked secret, and the `REVDEV_SPAWN_CONFINEMENT` deny), 40 pass. The non-regression tests (normal root beneath home, separator-safe siblings, `neverBoundSet` materialization) correctly stay green in both directions.

### Non-blocking notes (coverage only, not defects)

1. Spec §5.C item 11 (escape-hatch mode with an overlapping root still spawns) has no explicit unit test. It is correct by construction as verified above; a one-line test pinning it would prevent a future edit from accidentally hoisting the guard into the handler. Optional follow-up.
2. Spec §5.B item 8 (belt-and-suspenders: `buildConfinedEnv` output never contains the key) is moot in the real flow because `filterCallerEnv` throws before `buildConfinedEnv` ever receives it — the deny test covers the reachable path. Fine as-is.

### Disposition

Clears guardrail-2. **Crown-jewel-adjacent surface: the owner applies `sec-review:approved` and merges** — I do not self-merge or self-label. jv#422 (the lockstep spec pair) already merged and is consistent with the shipped code.

RevealUIStudio:
## Guardrail-2 verdict: APPROVE — independent re-derivation, post-merge (non-author, Fable 5)

Second, fully independent pass per the coordinator's request: the earlier verdict comment on this PR was treated as an untrusted first pass (its session identity was disputed), and everything below was re-derived from the code with no reliance on it. By the time this review ran, the owner had already applied `sec-review:approved` and merged (05:09:48Z, merge commit `b843f29`), so this verdict is a post-merge verification: had it dissented, the outcome would have been a repair PR, not a label.

### Independently verified against the design (grounded at `b843f29`)

1. **Refuse-over-mask, both overlap directions.** `assertGrantedRootBindable` throws on `repoReal === operatorHome`, `within(repoReal, operatorHome)` (ancestor), and for every never-bound entry on `within(repoReal, nb) || within(nb, repoReal)`. `within()` is the separator-safe shape (`target === root || target.startsWith(root + sep)`), and the non-regression tests pin the separator edge cases (`.sshkeep`, `/base/op-scratch`) plus the required asymmetry: a root beneath the home is NOT refused.
2. **Placement.** The guard is the first statement of `linuxBubblewrapBackend.spawnConfined()`, before any argv is assembled, co-located with the tmpfs-then-bind sequence that creates the hazard. The `ConfinementBackend` interface doc carries the MUST-validate contract for future backends.
3. **Escape hatch unguarded by construction, verified in the merged source.** `spawnConfined` has exactly one call site, inside the `conf.backend` branch of `agent.spawn`; the `conf.escapeHatch` branch sets `file = command; argv = args` directly (with the warn + `confinement='none'` receipt), and the third branch refuses outright. No assumption needed.
4. **Env deny.** `CALLER_ENV_DENY_EXACT` is checked FIRST in `filterCallerEnv`, before the exact-allow and the `REVDEV_` prefix allowance, and throws naming the key. Membership criterion (daemon-control keys that configure or weaken a boundary) is documented and correctly excludes the prune knobs. `buildConfinedEnv` routes caller env through `filterCallerEnv`, so the deny propagates at composition level.
5. **Never-bound set.** `neverBoundSet` materializes the parent spec's eight home-relative secrets + `/mnt/c` + `/mnt/e` + `/run/user/<uid>` (getuid-conditional) + `dataDir` (the jv#422 lockstep addition — spec text confirmed consistent), each in lexical + realpath-if-exists form, computed per spawn. The symlinked-secret case and the S1 `resolveOperatorHome` realpath are tested with on-disk fixtures.

### Red-first proof, re-run from scratch (detached worktree at `b843f29`)

- Baseline: `confinement.test.ts` **53/53**, full daemon suite **619/619 (29 files)**, `tsc --noEmit` clean. (The suite prints an unrelated "License validation failed: license expired" warning from a licensing fixture; nothing fails.)
- Neutered tree (predicate body stubbed to `return` + deny set emptied, one pass): **13 tests RED, 40 green** — the reds are exactly the new acceptance tests (home identity/ancestor, secret is/inside/contains, NTFS mounts, dataDir in and out of the home, symlinked secret, both backend-wiring refusals, and the env deny), and the non-regressions correctly stay green. Restored: 53/53 again.

### Relation to the prior comment

My read **agrees with the prior APPROVE on every technical point**, including its two non-blocking coverage notes, which I reproduce as my own findings: (1) design §5.C11 (escape-hatch spawn with an overlapping root) has no explicit pinning test — correct by construction today, and a one-line test would keep the guard from ever being hoisted into the handler; (2) §5.B8 is moot on the reachable path since `filterCallerEnv` throws upstream of `buildConfinedEnv`. Nothing in the prior comment was found to be wrong, inflated, or unverifiable; whatever its provenance, its content checks out against the code.

### Disposition

Clears guardrail-2 on my independent reading. The label and merge were the owner's actions and preceded this verdict; no further gate action is needed. This verdict ran on Fable end to end with no safeguard pause.

RevealUIStudio:
## Concurring verdict: APPROVE — third fully-independent pass, convergent (non-author, Fable 5)

A parallel Fable session (background job, distinct from the 05:35Z reviewer — coordination claim recorded on the fleet workboard at 05:27Z) ran the full re-derivation protocol concurrently and independently: diff grounded at PR head `426482d` against merge-parent `8ec236f`, every design claim re-verified from source, and the red-first proof re-run from scratch in its own detached worktree. The 05:35Z verdict landed first, so per dup-work discipline this is recorded as a concurrence, not a duplicate full verdict.

**Convergence:** every finding matches the 05:35Z comment — all four design claims (both overlap directions + separator-safe `within()` + the beneath-home asymmetry; guard first-statement placement; escape hatch unguarded by construction with exactly one `spawnConfined` call site; env deny checked first) verified identically, and the same two non-blocking test-plan deviations (§5.C11 unpinned escape-hatch/overlap case, §5.B8 moot-by-throw) reproduced independently.

**One increment of evidence the single-neuter run doesn't show — the mutation matrix was run as THREE separate mutations, each restored before the next:**

| Mutation | Red | What it proves |
|---|---|---|
| Guard CALL stashed in `spawnConfined` (predicate intact) | **2** (both backend wiring tests; predicate tests stay green) | the wiring tests pin the CALL SITE independently — a future edit that keeps the predicate but drops the call is caught |
| Predicate stubbed to a no-op | **12** (all ten refusal tests + both wiring tests) | the acceptance surface bites on the predicate itself |
| Env deny disabled | **1** (GAP-320b by-name rejection, incl. deny-beats-prefix ordering) | the deny check is load-bearing, not shadowed by the prefix allowance |

Restored baseline each time: 53/53. Full daemon suite **619/619 (29 files)**, `tsc --noEmit` clean, real-bwrap integration **8/8** on bubblewrap 0.9.0 including the GAP-320a §5.D refusal alongside the passing DAILY DRIVER happy path. jv#422 spec lockstep confirmed consistent. Ran on Fable end to end, no safeguard pause. Worktree `~/revfleet/.wt/gap320-rederive-bg` removed after the run.

**Verdict: APPROVE — concurring.** Three independent passes (the disputed 05:17Z first pass, the 05:35Z re-derivation, and this one) now converge on the same conclusion and the same two optional follow-ups; the merged state stands.

PR #280, opened 2026-07-11
fix(daemon): close file-layer never-bound reachability at project.open + resolution (GAP-326)

RevealUIStudio:
## What

Closes the file-layer never-bound reachability gap (GAP-326, design in the private
tracker `docs/gap-specs/GAP-326-file-layer-never-bound-reachability-design.md`).
A client-owned verified signer could `project.open` a secret-bearing tree and read
secrets straight through `file.read` with no `agent.spawn` involved, so the
confinement never-bound guard never ran. This lands all four parts as one
invariant (splitting reopens the fail-open windows between them):

- **D1 registration gate** (`project.open`): rejects a non-repo root (making the
  long-standing comment true) and refuses a root overlapping the never-bound set
  in either direction, or the operator home, before any registration side effect.
- **D2 resolution belt** (shared target resolver): refuses a resolved target that
  lands on the never-bound set. One precomputed prefix scan, no added I/O.
- **D3 startup re-validation** (`restoreProjectRoots`): re-checks every persisted
  row and evicts (deletes the row + logs a loud line) any that now fails, so
  never-bound-set growth is retroactively protective.
- **One list, imported**: `confinement.ts`'s `neverBoundSet` plus a new shared
  `findNeverBoundOverlap` predicate are consumed by both the spawn side and the
  file side. No second literal list. The file side names the offending entry by
  CLASS ("ssh keys"), never the full secret path, so the error is not an oracle
  back to a possibly hostile caller.

Registration-fixture scratch dirs that call `project.open` now carry a `git init`
line, per the design's audit item (the non-repo rejection would otherwise break
them).

## Never-bound set is imported, not copied

`filegit.ts` imports `neverBoundSet`, `findNeverBoundOverlap`, and
`resolveOperatorHome` from `confinement.ts`. `assertGrantedRootBindable` (spawn
side) was refactored to consume the same `findNeverBoundOverlap` predicate. A grep
for a second never-bound literal list finds none.

## Prove-red evidence (stashed-fix red run)

Stashed only the two source files (`confinement.ts` + `filegit.ts`), leaving the
tests, and ran the seam-free end-to-end attack file against the pre-fix daemon:

```
BEFORE (source stashed to base):
  × REFUSES the never-bound attack ... (project.open of a secret-bearing tree)
      AssertionError: expected undefined to be defined   <- open.error is undefined
                                                            = project.open SUCCEEDED
                                                              = the attack works
  × REFUSES a non-repo root at project.open
      AssertionError: expected undefined to be defined   <- bare dir registered
  ✓ leaves a legit repo root unaffected (regression control stays green)
  Tests  2 failed | 1 passed (3)

AFTER (fix restored):
  Tests  3 passed (3)
```

The CI prove-red gate agrees: reverting the source to base, 2 changed TypeScript
test units are red against base ("all active languages carry failing-first
evidence").

## Tests

- New `filegit-neverbound.test.ts` (real signed socket, no seams): the prove-red
  attack + non-repo refusal + a legit-repo regression control.
- New `filegit-neverbound-unit.test.ts`: the D1/D3 refusal matrix (operator home,
  root == / inside / containing a secret, separator-safe negatives, class-only
  error), the D2 belt in isolation (a pre-fix overlapping root seeded directly),
  and D3 eviction (`restoreProjectRoots` evicts overlapping + non-repo rows, keeps
  the clean one, logs each).
- Full daemon suite green: 31 files, 634 tests (was 619). Adversarial-isolation
  (25 cases), `spawn.test.ts`, and the confinement suites stay green. Typecheck,
  Biome, private-leak scan, and the zero-9P fence all pass.

## Review

This is a crown-jewel-adjacent security surface. Per the review-before-merge
convention, a **guardrail-2 non-author verdict is required before merge**; the
author does not self-approve and does not merge. Owner disposes.

### DIFF
--- a/packages/daemon/src/__tests__/filegit-enrollment.test.ts
+++ b/packages/daemon/src/__tests__/filegit-enrollment.test.ts
@@ -14,6 +14,7 @@
  * pointed at a fixture (production writes it root-owned at install).
  */
 
+import { execFileSync } from 'node:child_process';
 import { mkdtemp, rm, writeFile } from 'node:fs/promises';
 import { connect, type Socket } from 'node:net';
 import { tmpdir } from 'node:os';
@@ -119,6 +120,9 @@ describe('daemon enrollment gate + per-agent root scoping', () => {
     dataDir = await mkdtemp(join(tmpdir(), 'revdev-enroll-'));
     repoA = await mkdtemp(join(tmpdir(), 'revdev-enroll-repoA-'));
     repoB = await mkdtemp(join(tmpdir(), 'revdev-enroll-repoB-'));
+    // project.open now rejects a non-repo root (GAP-326 D1); make the fixtures repos.
+    execFileSync('git', ['init', '-q', '-b', 'main'], { cwd: repoA });
+    execFileSync('git', ['init', '-q', '-b', 'main'], { cwd: repoB });
     socketPath = join(dataDir, 'harness.sock');
     const anchor = join(dataDir, 'trusted-client-fingerprint');
     // Anchor trusts the (agentId, fingerprint) PAIRS of A and B but NOT evil.
--- a/packages/daemon/src/__tests__/filegit-neverbound-unit.test.ts
+++ b/packages/daemon/src/__tests__/filegit-neverbound-unit.test.ts
@@ -1,0 +1,227 @@
+/**
+ * GAP-326 — file-layer never-bound guards, UNIT coverage via the @internal
+ * seams (no socket): the D1/D3 registration-refusal matrix, the D2 resolution
+ * belt in isolation (a pre-fix persisted overlapping root seeded directly), and
+ * the D3 startup eviction of a persisted overlapping row.
+ *
+ * The end-to-end attack + regression live in filegit-neverbound.test.ts (real
+ * socket, no seams). This file drives the individual guards where a seam is the
+ * cheapest faithful surface.
+ */
+
+import { execFileSync } from 'node:child_process';
+import { mkdir, mkdtemp, realpath, rm, stat, writeFile } from 'node:fs/promises';
+import { tmpdir } from 'node:os';
+import { join } from 'node:path';
+import { PGlite } from '@electric-sql/pglite';
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
+
+// Capture the eviction log line (D3) without a real logger. Only createLogger is
+// stubbed; every other logger export keeps its real implementation.
+const { warnCalls } = vi.hoisted(() => ({
+  warnCalls: [] as Array<{ msg: string; meta: unknown }>,
+}));
+vi.mock('@revealui/utils/logger', async (importActual) => {
+  const actual = await importActual<typeof import('@revealui/utils/logger')>();
+  const stub = (): Record<string, unknown> => ({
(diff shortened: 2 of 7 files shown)


RevealUIStudio:
## Guardrail-2 verdict (Fable, non-author, adversarial) — APPROVE [CONFIRMED]

The GAP-326 file-layer never-bound reachability hole is closed exactly to the design contract (jv#448). Reviewed adversarially and verified by independent execution, not just diff-reading.

**Correctness of the three enforcement points:**
- **D1 (project.open):** rejects non-repo roots (making the long-drifted `filegit.ts:509` comment true AND making `assertNoExecConfig` unconditional — closes both gaps at once), and `assertRootAvoidsNeverBound(real)` refuses a root overlapping the never-bound set in either direction or the operator home, BEFORE any registration side effect.
- **D2 (resolution belt):** `assertTargetAvoidsNeverBound` is applied in BOTH `resolveInRoot` return paths — the mustExist realpath'd-target branch (read/delete/stat) and the parent-realpath'd write/create branch. One precomputed prefix scan, no added I/O.
- **D3 (startup re-validation):** `restoreProjectRoots` re-applies D1 to each persisted row and evicts (DELETE + `log.warn` naming the root) any non-repo or overlapping row, so never-bound-set growth is retroactively protective against pre-fix rows.

**Surface coverage — no bypass (independently checked):** every `file.*` handler (read/write/delete/stat) routes through `requireRoot`→`resolveInRoot` (now D2-gated); every `git.*` handler runs `runGit([...], repoReal)` on a root D1 guarantees is a registered repo off the never-bound set, and git confines pathspecs to the repo. Symlink escape is still caught earlier by the realpath+`within()` descendant check (a symlink to ~/.ssh resolves out-of-root and is rejected before D2). Hardlink inode residue remains explicitly out of scope (Layer-B, documented).

**One-list invariant HELD:** grep confirms the only literal never-bound lists (`NEVER_BOUND_HOME_RELATIVE` / `NEVER_BOUND_ABSOLUTE`) live in `confinement.ts`; `filegit.ts` imports `neverBoundSet` + the new shared `findNeverBoundOverlap`, and `assertGrantedRootBindable` (spawn) was refactored to consume the same predicate. No copied logic.

**Oracle-safety:** errors handed to a possibly-hostile `project.open`/`file.*` caller name only the CLASS ("ssh keys"), never the full home path — while the spawn side keeps the full-path error (operator-known). This is why the shared predicate is `findNeverBoundOverlap` rather than the design's illustrative single `assertPathAvoidsNeverBound`; the deviation is correct and the design offered that name as an "e.g." — the hard requirements (one set, one shared predicate, both consume) are all met.

**Independently executed:**
- Full daemon suite **634/634 green** (was 619; +15 across the two new test files), run from the branch worktree.
- **Prove-red reproduced myself** (checked out `filegit.ts`+`confinement.ts` at `origin/test`, kept the fix's tests): attack suite **2 failed / 1 passed** — `project.open` of a secret-bearing tree SUCCEEDS and bare-dir registration succeeds against base. Restored the fix → **3 passed**. The attack is real and the fix closes it.

Crown-jewel-adjacent surface, clean. This is the last load-bearing RevDev Part B item. Owner disposes (gate clears on `sec-review:approved`).

PR #285, opened 2026-07-16
feat(daemon): contract-test RPC_METHODS against the handler registry

RevealUIStudio:
## Problem

`RPC_METHODS` (protocol), the daemon handler registry, and Studio's `HARNESS_RPC_MAP` were three hand-maintained lists that had drifted apart. `RPC_METHODS` declared seven methods that no handler served and that nothing called:

- `harness.list`, `harness.execute`, `harness.info`, `harness.listRunning`, `harness.syncConfig`, `harness.diffConfig`
- `session.history`

It simultaneously omitted whole registered surfaces: `file.*`, `git.*`, `project.*`, `inference.chat`, `inference.generate`, and `identity.rotate`. The constant's own header claimed "every method the daemon exposes is listed here", which was false.

Separately, `merge.update` was registered and validated in the daemon but had zero callers, and the bridge (the MCP server for external tools) wrapped `merge.request/status/list` but not `merge.update`.

## Changes

- **Introspectable registry**: `listRegisteredMethods()` exported from the daemon returns the sorted registered method names at runtime.
- **Honest constant**: `RPC_METHODS` regenerated to exactly the 67 registered methods (seven phantoms dropped, the omitted surfaces added). Its header now states it is contract-tested against the daemon registry, not hand-maintained.
- **Contract test (daemon)**: `rpc-contract.test.ts` registers all handlers exactly as production startup does (importing the index side-effect graph, so `spawn.ts` overwrites the `agent.ts` desktop-only stubs) and asserts `listRegisteredMethods()` equals `Object.values(RPC_METHODS)` in both directions, with a sorted diff in the failure message.
- **Contract test (Studio)**: `HARNESS_RPC_MAP` is now exported, and the Studio invoke test asserts every value in it is a member of `RPC_METHODS`.
- **Bridge**: new `merge_update` MCP tool mirroring `merge_request/status/list`, giving the `merge.update` handler a real consumer. The daemon handler is unchanged.

## Deviation from brief (with rationale)

The brief asked for the Studio-map assertion to live inside the daemon contract test, importing `apps/studio/src/lib/invoke.ts`. That is not viable without breaking the build: the daemon's `tsc --emitDeclarationOnly` step compiles `src/**` (including `__tests__`) under `rootDir: "src"`, and the daemon has no dependency on the `studio` app. Importing an `apps/studio` file from a daemon test fails declaration emit (TS6059, file outside rootDir) and inverts the package to app dependency direction. The invariant "Studio's map only references real RPC methods" is owned by Studio, which already has an `invoke.test.ts` and a vitest setup. The assertion therefore lives there, still importing the map (no text parsing) and asserting membership against `RPC_METHODS`. This required adding `@revdev/protocol` as a Studio devDependency (test-only).

## Verification

- `pnpm -r --filter=!studio build` green (daemon `tsc` declaration step compiles the new test clean)
- `pnpm --filter @revdev/daemon test`: 32 files, 635 tests pass (includes the new contract test)
- `pnpm --filter @revdev/bridge test`: 5 pass
- Studio invoke test: 29 pass (includes the new `HARNESS_RPC_MAP` contract)
- `pnpm typecheck:studio` green (after generating ts-rs bindings)
- Biome clean on all touched files
- Prove-red: reintroducing a phantom and dropping a real method both turn the contract test red with a precise diff

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -72,6 +72,9 @@ jobs:
       # #143 failure mode) then `vite build`, mirroring build-windows.ps1's frontend
       # step, catching type errors AND bundling errors on every PR.
       - run: bash apps/studio/scripts/generate-types.sh
+      # Studio's contract test imports RPC_METHODS from @revdev/protocol, a
+      # workspace dep resolved from its dist output, so build it first.
+      - run: pnpm --filter "@revdev/protocol" build
       - run: pnpm --filter studio build
 
   test:
--- a/apps/studio/package.json
+++ b/apps/studio/package.json
@@ -29,6 +29,7 @@
     "react-dom": "^19.2.7"
   },
   "devDependencies": {
+    "@revdev/protocol": "workspace:*",
     "@tailwindcss/vite": "^4.3.2",
     "@tauri-apps/cli": "^2.11.4",
     "@testing-library/jest-dom": "^6.9.1",
--- a/apps/studio/src/__tests__/lib/invoke.test.ts
+++ b/apps/studio/src/__tests__/lib/invoke.test.ts
@@ -231,3 +231,21 @@ describe('invoke bridge (Tauri mode)', () => {
     expect(result.wsl_running).toBe(true);
   });
 });
+
+describe('HARNESS_RPC_MAP contract', () => {
+  it('routes every command to a method the daemon actually registers', async () => {
+    const { HARNESS_RPC_MAP } = await import('../../lib/invoke');
+    const { RPC_METHODS } = await import('@revdev/protocol');
+    const known = new Set<string>(Object.values(RPC_METHODS));
+
+    const unknownTargets = Object.entries(HARNESS_RPC_MAP)
+      .filter(([, method]) => !known.has(method))
+      .map(([command, method]) => `${command} -> ${method}`)
+      .sort();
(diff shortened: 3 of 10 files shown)


RevealUIStudio:
## Review (fable, adversarial pass)

**Checked:**
- The load-bearing security question: whether RPC_METHODS acts as an exposure gate anywhere. Verified it does not: its only consumer is bridge's name-lookup table, and method gating lives at the dispatch layer (MUTATING_OR_CONTENT_METHODS + signature requirements + the license guard), so adding the 26 previously-omitted entries widens nothing.
- The daemon contract test imports the ../index.js side-effect graph (handler registration incl. the spawn-over-stubs ordering) without calling startDaemon, so no socket or storage is touched in tests.
- The Studio-side membership test imports the real map (no text parsing); placing it in the Studio package instead of the daemon is the correct dependency direction and the stated deviation is sound.
- bridge merge_update mirrors its three siblings and inherits the same daemon-side validation, identity gate, and license tier as the rest of the merge family.
- Prove-red was demonstrated in both directions (reintroduced phantom and dropped real method both fail with a precise diff).

**Findings:** none.

**Verdict: APPROVE.** Not a gated surface per the fleet checker; merge on green is the owner's call.

PR #286, opened 2026-07-16
feat(daemon): native security tool-guard for tool actions

RevealUIStudio:
## What

Adds a RevDev-native security pattern guard on the daemon request path, the native sibling of the Claude Code PreToolUse pattern scanner. Pattern data now lives in exactly one canonical manifest with two consumers: the daemon (this PR) and the PreToolUse hook (a vendored copy, changed in the private hooks repo).

## How

- **One manifest, typed loader.** `packages/daemon/src/tool-guard/patterns.json` (versioned) carries dangerous-command matchers, credential-path rules, blocked-write prefixes, protected env/lock file rules, and secret-content indicators. `patterns.ts` validates its shape at load and computes a stable content hash. Zero regex per the no-regex rule: substring/prefix data plus named character-boundary predicates.
- **Pure guard.** `tool-guard/index.ts` exposes `evaluateToolAction(action)` returning an allow/deny verdict for `command`, `read`, `write`, and `delete`. Wired into `file.read` / `file.write` / `file.delete` (path + content checks) and `agent.spawn` (command-class check that **composes with, never replaces**, the existing root-scoping + confinement).
- **Denials.** A distinct JSON-RPC error code (separate from the license guard's) plus an `events` row carrying the rule id.
- **Fail closed at startup.** `initToolGuard()` runs before the socket binds; an unloadable manifest refuses to start, mirroring the license guard.
- **Vendor sync + lockstep.** `sync-vendored.ts` writes the hook's vendored copy and prints the hash. A unit test pins `(version, hash)` so a pattern edit that forgets the version bump fails the build.

## Scope honesty

The guard covers what flows **through the daemon**. PTY-internal tool calls a foreign harness makes are out of scope (the guard is a pure function so the same entry point applies if the native runtime later gains its own tool loop). `git.*` handlers are intentionally not path-guarded: they are structured and root-confined, and a secret-path check there would regress legitimate repo files (e.g. a project's own `.npmrc`).

## Tests

Per-category block + allow (command/read/write/delete), a manifest-corruption refuses-to-start test, `validateManifest` corruption cases, matcher word-boundary fidelity, secret-token fidelity, and the hash/version lockstep. Full `@revdev/daemon` suite green (665 tests). `pnpm lint` and `pnpm typecheck` clean.

## Security review

This PR touches daemon security surfaces (spawn + file I/O request path). **It requires a recorded security verdict before merge** per the review-before-merge gate; do not self-merge.

### DIFF
--- a/.leakignore
+++ b/.leakignore
@@ -30,3 +30,9 @@ docs/INDEX.md                                              private-jv-repo
 # issue bodies. This is the config that enforces the rule, not a leak of
 # it (mirrors the ci.yml anti-regression entries).
 .gitleaks.issues.toml                   lit-username,abs-home-path,private-jv-repo,private-jv-name
+
+# The native tool-guard manifest carries external-drive mount prefixes as
+# blocked-WRITE-path data (defensive, extracted verbatim from the Claude Code
+# PreToolUse hook): they tell the guard to REFUSE writes under those mounts.
+# This is a denylist entry, not a reference to a real drive location.
+packages/daemon/src/tool-guard/patterns.json              lts-drive
--- a/.leakignore
+++ b/.leakignore
@@ -30,9 +30,3 @@ docs/INDEX.md                                              private-jv-repo
 # issue bodies. This is the config that enforces the rule, not a leak of
 # it (mirrors the ci.yml anti-regression entries).
 .gitleaks.issues.toml                   lit-username,abs-home-path,private-jv-repo,private-jv-name
-
-# The native tool-guard manifest carries external-drive mount prefixes as
-# blocked-WRITE-path data (defensive, extracted verbatim from the Claude Code
-# PreToolUse hook): they tell the guard to REFUSE writes under those mounts.
-# This is a denylist entry, not a reference to a real drive location.
-packages/daemon/src/tool-guard/patterns.json              lts-drive
--- a/packages/daemon/package.json
+++ b/packages/daemon/package.json
@@ -39,6 +39,7 @@
     "dev": "tsup --watch",
     "start": "node dist/cli.js",
     "start:detached": "node dist/cli.js --detach",
+    "sync:tool-guard": "node dist/tool-guard/sync-vendored.js",
     "setup:systemd": "bash systemd/install.sh",
     "db:generate": "drizzle-kit generate",
     "db:push": "drizzle-kit push",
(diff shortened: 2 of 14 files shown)


RevealUIStudio:
## Security review (fable/opus adversarial pass, gated surface)

Reviewed against the design spec and the pre-PR daemon behavior. This is a guardrail-2 surface (filegit.ts, spawn.ts), so the load-bearing question was: does the guard weaken or bypass any existing authorization?

**Answer: no. It is strict defense-in-depth.**
- `evaluateToolAction` is pure and synchronous and only ever returns a DENY or ALLOW; it never grants. It is layered AFTER the existing gates at every call site: `file.*` evaluate on the already `resolveInRoot`-ed target (the real path, not user input), and `agent.spawn` checks the joined command line before any process is forked, composing with (not replacing) the sandbox confinement.
- Fail-closed at startup: `initToolGuard` runs before the socket binds and throws on an unloadable/invalid manifest, aborting with nothing to tear down. `loadPatterns` caches after first load, so a boot that succeeds can never later fail-open at evaluation time.
- Denials carry a distinct JSON-RPC code (-32006, not the license guard's -32001) and log a best-effort `tool-guard.denied` events row; a failed insert never masks the denial.

**Adversarial checks that passed:**
- No per-call file IO (manifest cached); no way to swap the manifest at runtime.
- Path predicates operate on forward-slash-normalized paths; credential/blocked-write/env/lock checks are endsWith/includes/startsWith + named predicates, zero regex, and the `.env` committed-template exemption is preserved (isEnvTemplate).
- The 48-fixture pre/post hook parity battery shows byte-identical exit codes and first-stderr-line across every extracted category, so the Claude-side refactor is behavior-preserving; the hook fails OPEN (correct for a session guard) while the daemon fails CLOSED (correct for a service).
- `git.*` intentionally not path-guarded: those handlers are structured and root-confined, and a credential-path check there would false-positive on legitimate in-repo files (a project's own `.npmrc`). Sound scoping call, documented.
- PTY-internal tool calls remain out of scope by design; `evaluateToolAction`'s pure-function shape is the future hook point if the native harness gains its own tool loop.

**Course-correction confirmed landed:** `/mnt/e/` (the operator's machine-specific LTS mount) was removed from the public `patterns.json` and retained only as a Claude-side `LOCAL_BLOCKED_WRITE_PREFIXES` entry, with a daemon test asserting the mount does not ship. The eight `/home/...` test-fixture leak-scan hits were rerooted to `/work/repo`. Private Leak Scan and all scanners now pass.

**Findings: none blocking.** One note for a future iteration (not a merge blocker): the documented regex-to-predicate micro-widenings on non-realistic inputs are acceptable because they only ever widen toward MORE blocking, never less; exit-code behavior is identical on every realistic payload.

**Verdict: APPROVE.** Clearing the gate (the `sec-review:approved` label) and the merge are the owner's actions.

PR #287, opened 2026-07-16
feat(daemon): wire browser-mode agent list/remove and ollama rpc

RevealUIStudio:
## What

Studio's browser mode (remote daemon over HTTP) previously rejected the agent-list/remove and Ollama inference commands as desktop-only, because the daemon had no matching RPC. This wires the commands the daemon can actually serve and adapts result shapes back to the Studio types, and leaves the genuinely host-only ones rejecting.

## Per-family decision

| Command family | Decision | Why |
|---|---|---|
| `agent.list` / `agent.remove` | **Wired** | The daemon owns a real process registry (`agent_processes` + the live PTY map) but had no way to enumerate or prune it over RPC. `agent.list` returns the caller's processes; `agent.remove` kills the live PTY then deletes the row (output cascades). Both are signature-required + Pro, exactly like the rest of `agent.*`. |
| `inference_ollama_status` / `_models` / `_pull` | **Wired** | The daemon's `inference.*` handlers ARE the Ollama HTTP integration. Status and models both come from `inference.status` (it already returns the model list, so no new "models" method); pull maps to `inference.pull`. Shapes are adapted at the Studio boundary. |
| `inference_ollama_delete` | **Wired (new daemon method)** | Ollama supports `DELETE /api/delete`. Added `inference.delete` as thin Max-tier model management alongside pull/start/stop. |
| `inference_ollama_start` / `_stop` | **Kept desktop-only** | These manage the Ollama **server** lifecycle (`ollama serve` / `pkill`). The daemon only speaks HTTP to an already-running server, so it cannot start or stop a host process. (The daemon's `inference.start`/`stop` are model warm/unload, a different op.) |
| `inference_snap_*` | **Kept desktop-only** | Host snap package management (`sudo snap install/remove`). No daemon equivalent, and the daemon does not manage host packages. |

## New daemon methods

- `agent.list` — enumerate the caller's `agent_processes` (owner-scoped), signature-required, Pro.
- `agent.remove` — stop-if-live + prune the row, signature-required, Pro.
- `inference.delete` — `DELETE /api/delete`, Max-tier (in `METHOD_MIN_TIER`), identity-exempt like its inference siblings.

All three were added to the protocol `RPC_METHODS`, the validation schemas, and the tier-classification set in lockstep. `agent.list`/`agent.remove` were mirrored into `server.ts` `MUTATING_OR_CONTENT_METHODS`, the `signature-gate` lock test, and the Rust `signing.rs` `requires_signature` set (+ its test).

The RPC contract test (`packages/daemon/src/__tests__/rpc-contract.test.ts`) asserts the daemon handler registry EXACTLY equals `RPC_METHODS`, so any new handler must be mirrored in `packages/protocol/src/methods.ts` or the build fails; that mirroring is included here.

## Shape adaptation

Browser mode returns the same types the Tauri path returns. Adapters in `apps/studio/src/lib/invoke.ts` map:
- `inference.status` -> `OllamaStatus { installed, running, version }` (`installed` tracks `running`: the daemon can only confirm install when Ollama answers over HTTP).
- `inference.status.models` -> `OllamaModel[]` (with a formatted size string).
- `inference.pull` -> `ModelPullResult { success, message }` (and `modelName` -> `model` param rename for the daemon schema).
- `agent.list` rows -> `AgentSessionInfo[]`. The daemon's `agent.spawn` is a generic command spawner with no inference backend/model/prompt, so `name` is the command and `backend` is `null`. `AgentSessionInfo.backend` was widened to `AgentBackend | null`, and SpawnerPanel labels a null backend as `PTY`.

## Tests

- Daemon: `agent.list` (owner-scoped, unsigned rejected) and `agent.remove` (stop+prune, cross-agent rejection, unknown id) in `spawn.test.ts`.
- Studio: browser-mode adapter integration tests (status/models/pull/agent-list) driving the real `invoke()` path with a mocked fetch, plus the existing `HARNESS_RPC_MAP` contract block.

## Verification

- `pnpm -r --filter=!studio build` — pass
- `pnpm --filter "@revdev/daemon" test` — 676 pass (incl. rpc-contract, guard tier-classification, signature-gate lock)
- `pnpm --filter "@revdev/protocol" build` — pass
- `pnpm typecheck:studio` — pass
- `pnpm --filter studio test` — 546 pass
- Biome clean on touched files

Rust (`signing.rs`) was not compiled locally (no cargo in this pass); the change is a two-arm match addition mirrored in its unit test, and CI compiles Rust via the typecheck-studio job's `generate-types.sh` (`cargo test --lib`).

### DIFF
--- a/apps/studio/src-tauri/src/signing.rs
+++ b/apps/studio/src-tauri/src/signing.rs
@@ -84,6 +84,11 @@ pub fn requires_signature(method: &str) -> bool {
             | "agent.input"
             | "agent.resize"
             | "agent.output"
+            // agent.list returns another-agent-invisible process metadata and
+            // agent.remove kills + prunes a process; both are self-scoped to the
+            // verified signer. MUST mirror server.ts MUTATING_OR_CONTENT_METHODS.
+            | "agent.list"
+            | "agent.remove"
             // session.end evicts the target's project roots and kills its PTYs.
             // Signature-required so the daemon can self-scope it to the verified
             // signer instead of a caller-supplied `sessionId`.
@@ -527,6 +532,8 @@ mod tests {
         assert!(requires_signature("agent.input"));
         assert!(requires_signature("agent.resize"));
         assert!(requires_signature("agent.output"));
+        assert!(requires_signature("agent.list"));
+        assert!(requires_signature("agent.remove"));
         assert!(requires_signature("session.end"));
         // GAP-312: same eviction primitive as session.end, fleet-wide fan-out.
         assert!(requires_signature("harness.prune"));
--- a/apps/studio/src/__tests__/lib/invoke.test.ts
+++ b/apps/studio/src/__tests__/lib/invoke.test.ts
@@ -249,3 +249,107 @@ describe('HARNESS_RPC_MAP contract', () => {
     ).toEqual([]);
   });
 });
+
+// Browser mode routes the newly-wired commands over the daemon HTTP gateway and
+// adapts each daemon result back to the Studio type. These drive the real
+// invoke() path with a mocked fetch so the mapping, param rename, and result
+// adapters are all exercised end to end.
+describe('browser-mode daemon adapters', () => {
+  const DAEMON_URL = 'https://daemon.test';
+
+  /** A fetch stub returning a JSON-RPC success envelope wrapping `result`. */
+  function rpcOk(result: unknown): ReturnType<typeof vi.fn> {
+    return vi.fn().mockResolvedValue({
+      ok: true,
+      status: 200,
+      json: async () => ({ jsonrpc: '2.0', id: 1, result }),
+    } as unknown as Response);
+  }
+
+  beforeEach(() => {
+    vi.clearAllMocks();
+    delete (window as unknown as Record<string, unknown>).__TAURI_INTERNALS__;
+    localStorage.setItem('revdev-daemon-url', DAEMON_URL);
+    localStorage.setItem('revdev-daemon-token', 'test-token');
+  });
+
+  afterEach(() => {
+    localStorage.clear();
+    vi.unstubAllGlobals();
+    vi.restoreAllMocks();
+  });
+
+  it('adapts inference.status into OllamaStatus', async () => {
+    vi.stubGlobal(
+      'fetch',
+      rpcOk({ running: true, version: '0.1.30', url: 'http://localhost:11434', models: [] }),
+    );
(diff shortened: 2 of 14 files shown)


RevealUIStudio:
## Security review (fable/opus adversarial pass, gated surface)

Gated because this adds two new agent-spawn-surface RPC methods plus signature-gate and tier wiring. Load-bearing question: can the owner-scoping on agent.list/agent.remove be spoofed?

**Answer: no.** The scoping is anchored to a cryptographically verified principal, layered across both languages:
- `requireAgent` (spawn.ts:179) returns `ctx.agentId` ONLY when `ctx.boundVia === 'signature'`; it ignores params entirely (the `_params` arg is unused, kept for call-site parity) and throws on any unsigned or param-bound caller. So `owner_agent = ctx.agentId` is the verified Ed25519 signer, not caller-supplied.
- Both methods are in the daemon dispatch gate (`server.ts` MUTATING_OR_CONTENT_METHODS) AND in the Rust `signing.rs requires_signature()`, with a unit test asserting both. An unsigned request never reaches the handler.
- `agent.list` queries `WHERE owner_agent = $1` (caller only). `agent.remove` verifies ownership from the live registry when the process is running (`entry.ownerAgentId !== callerAgentId` throws) and from the persisted row otherwise (`r.owner_agent !== callerAgentId` throws) BEFORE the DELETE. No path deletes another agent's process; `agent_process_output` cascades by FK.

**Adversarial checks that passed:**
- No TOCTOU on remove: every code path that reaches the DELETE has already checked ownership against the verified signer.
- The per-family decision is correct: ollama server start/stop and snap.* stay desktop-only because the daemon only speaks HTTP to an already-running Ollama and does not manage host processes/packages. Wiring those would have meant giving a remote daemon host-process control, correctly avoided.
- The RPC contract test keeps RPC_METHODS == the registry, so these methods cannot silently drift out of the signature or tier maps.

**One noted observation (not blocking):** `inference.delete` is identity-exempt, matching its inference.pull/start/stop siblings. It is destructive (removes a downloaded Ollama model), so the whole inference surface trusts the local socket as unauthenticated. That trust-boundary decision predates this PR and is consistent; flagging only for awareness in case the inference surface is ever exposed beyond a trusted local socket, at which point delete (and pull/start/stop) would want the identity gate.

**Findings: none blocking. Verdict: APPROVE.** Clearing the gate (sec-review:approved) and merge are the owner's actions.

PR #288, opened 2026-07-16
feat(daemon): session-lifecycle advisory check surface

RevealUIStudio:
## What

Adds a native session-lifecycle advisory-check surface to the daemon: a check registry plus wiring into `session.register` that runs registered advisory checks warn-only and returns their warnings to the client.

This is the per-SESSION sibling of the per-ACTION `agent.spawn` confinement boundary. Where confinement gates a single action, this runs a set of registered ADVISORY checks when a session registers and hands their warnings back so Studio, Console, or any consuming harness can surface them. The daemon takes no action on the warnings.

## Registry API

`packages/daemon/src/session-checks/`

- `registerSessionCheck(name, fn)` where `fn: (ctx) => Promise<{ ok: boolean; warnings: string[] }>`. `ctx` carries the session `workDir` (project root) and `agentId`.
- `runSessionChecks(ctx)` runs every registered check and returns the collected warnings. Best-effort: each check runs inside its own try/catch, so a throwing check is logged and skipped and the run always resolves. Registration can never fail on a check error.
- `initSessionChecks()` registers the built-in checks; called once from `startDaemon`.
- `clearSessionChecks()` / `registeredSessionCheckNames()` support isolation and introspection.

`session.register` now runs the registered checks and adds a `warnings: string[]` field to its response. All existing response fields are unchanged. No new RPC method is added — the checks fire inside the existing handler, so the method contract is untouched.

## The ported check

A generic canonical **doc-locations** check. It takes a project root plus a config of path rules and carries no product-specific paths:

- `restrictedDirs` — directories whose direct children are restricted to an allowlist (misplaced files/subdirectories are flagged).
- `requiredFileInSubdirs` — parent directories whose immediate subdirectories must each contain a required index file.

It reads only the immediate children of the configured directories (no recursion, no traversal), reports warnings, and mutates nothing. The built-in registration ships an **empty** rule set, so the check is a no-op until a consuming project supplies its own canonical-path rules. This keeps the surface provider-agnostic.

### Deferred, on purpose

A doc-currency-style stale-fact check is intentionally NOT ported here. Its actual rules are deployment-specific stale facts that do not port verbatim to a customer project; porting it needs a customer-generic configurable-rule design first. No such facts are hardcoded into a shipped daemon check.

## Tests

- Registry/runner unit tests: a passing check contributes no warnings; a warning check surfaces its warnings; a throwing check is skipped and the run still succeeds; warnings collect across multiple checks.
- doc-locations unit tests: flags a misplaced file, an unexpected subdirectory, and a subdirectory missing its required file; passes a clean tree; empty config and absent directories are no-ops.
- Integration test: through a real daemon socket, `session.register` returns an empty `warnings` array on a clean run, and surfaces a warning check's message while a co-registered throwing check is skipped (registration still succeeds).

Prove-red confirmed: removing the response `warnings` field turns the integration test red.

## Verification

- `pnpm --filter "@revdev/daemon" test` — 690 passed (36 files), including the handler-contract coverage test.
- `pnpm -r --filter=!studio build` — clean.
- `pnpm --filter "@revdev/protocol" build` — clean.
- `pnpm typecheck:studio` — clean.
- Biome clean on touched files.

Constraints honored: TypeScript strict, ESM, zero authored regex (path joins + `Set`/array membership only), explicit return types on exports, no `any`, shared `@revealui/utils` logger, no new dependencies.

### DIFF
--- a/packages/daemon/src/__tests__/session-checks-integration.test.ts
+++ b/packages/daemon/src/__tests__/session-checks-integration.test.ts
@@ -1,0 +1,106 @@
+/**
+ * End-to-end proof that session.register runs the advisory checks and returns
+ * their warnings, without ever failing registration on a check error.
+ *
+ * Spins up a real daemon on a throwaway socket, registers a throwing check and
+ * a warning check AFTER startup (startDaemon's initSessionChecks already ran),
+ * then calls session.register over the wire.
+ *
+ * @vitest-environment node
+ */
+
+import { vi } from 'vitest';
+
+// Daemon startup can take >10s in CI (key generation + socket bind).
+vi.setConfig({ testTimeout: 30_000, hookTimeout: 30_000 });
+
+import { mkdtemp, rm } from 'node:fs/promises';
+import { connect, type Socket } from 'node:net';
+import { tmpdir } from 'node:os';
+import { join } from 'node:path';
+import { afterAll, beforeAll, describe, expect, it } from 'vitest';
+import { startDaemon } from '../server.js';
+import { clearSessionChecks, registerSessionCheck } from '../session-checks/index.js';
+
+function rpc(
+  socketPath: string,
+  method: string,
+  params: Record<string, unknown> = {},
+): Promise<unknown> {
+  return new Promise((resolve, reject) => {
+    const sock: Socket = connect(socketPath);
+    let buf = '';
+    const req = `${JSON.stringify({ jsonrpc: '2.0', id: 1, method, params })}\n`;
+    sock.on('connect', () => sock.write(req));
+    sock.on('data', (d) => {
+      buf += d.toString();
+      const nl = buf.indexOf('\n');
+      if (nl === -1) return;
+      const line = buf.slice(0, nl);
+      sock.end();
+      try {
+        const resp = JSON.parse(line);
+        if (resp.error) reject(new Error(`${resp.error.code}: ${resp.error.message}`));
+        else resolve(resp.result);
+      } catch (e) {
+        reject(e instanceof Error ? e : new Error(String(e)));
+      }
+    });
+    sock.on('error', reject);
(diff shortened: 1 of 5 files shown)


RevealUIStudio:
## Review (fable/opus)

Not a gated surface per the fleet checker (session-checks/ + the session.register additions match no sensitive path), but it wires into session.register, so a correctness + safety pass:

**Error isolation — correct.** runSessionChecks wraps every check in its own try/catch and always resolves (never rejects); a throwing check is logged and skipped. session.register cannot fail on a check error, which is the whole invariant. Verified in index.ts:235-250.

**Filesystem safety — correct.** doc-locations reads only immediate children (non-recursive readdir), treats absent/unreadable dirs as empty (not a violation), never mutates, never executes. Zero regex; Set/array membership only.

**Provider-agnostic — correct and as-scoped.** Default config is empty so the shipped check is a no-op; no RevFleet paths baked in. The doc-currency port was correctly deferred (its stale-fact rules are RevFleet-specific and need a customer-generic rule design first).

**Contract test stays green:** no new RPC method (checks fire inside session.register), and CI is green including rpc-contract.test.ts. (Note for the record: the PR body says rpc-contract.test.ts and tool-guard/ "don't exist" — they do, on this branch's base; the claim is wrong but harmless since nothing here adds a method. The chosen module pattern mirroring confinement.ts is fine.)

**One hardening nit (non-blocking):** createDocLocationsCheck does `join(root, rule.dir)` with no containment assertion, so an operator config with a `../` dir would read outside the project root. Not a vulnerability today (config is operator-supplied and ships empty; the check is read-only + warn-only), but when projects start supplying configs, a `within(root, resolved)` guard (the daemon already has one in confinement.ts) is cheap insurance. Fine to land as-is and add later.

**Verdict: APPROVE.** Not gated; merge on green is the owner's call.

PR #289, opened 2026-07-17
feat(daemon): add dup-work-claim check, harden doc-locations paths

RevealUIStudio:
## Summary

Extends the session-lifecycle advisory-check surface (added in #288) with one hardening fix and one new advisory check. Both run inside `session.register` and surface warnings to the client; neither adds an RPC method, blocks registration, or throws.

### 1. Harden `doc-locations` against path escape

The config-supplied `dir` / `parentDir` were joined to the project root with no containment check, so an operator config carrying a `../` segment would read outside the project root. The check now resolves the joined path and, if it escapes root (separator-safe containment, mirroring the `within()` helper in `confinement.ts`), skips that rule and emits a single warning naming the offending rule instead of reading outside the tree. Still read-only and warn-only.

### 2. New `dup-work-claim` advisory check

The provider-agnostic sibling of the fleet's duplicate-work-claim detector. When the same configured claim identifier appears in two or more git worktrees of a project, that is the visible signal that two sessions are building the same thing (the signal is already in `git worktree list`, it just needs reading before a second worktree is opened).

- **Config:** `{ claimMarkers?: string[] }` (e.g. `["TASK-", "GAP-"]`). A claim identifier is a marker followed by a run of alphanumeric characters. **Empty or unset makes the check a no-op**, so nothing about any one tracker's id scheme is baked into the daemon. Registered by default with an empty config, exactly like `doc-locations`.
- Extracts identifiers from each worktree's path basename and branch name (substring scan plus a character walk, zero regex).
- Runs `git worktree list --porcelain` via the daemon's existing hardened `runGit` helper (lazily imported to avoid a module load-time cycle: `server.ts` imports the session-check registry before its own handler map initializes, and `vcs.ts` registers handlers at module top-level).
- Best-effort: any git error yields a clean, no-warning result and the check never throws.

## Tests

- `doc-locations`: a rule with `dir: '../outside'` / `parentDir: '../outside'` produces the skip warning and reads nothing outside root (proven-red against the base without the guard).
- `dup-work-claim`: two worktrees claiming the same marker warn (real temp git repo + `git worktree add`); distinct markers stay clean; unconfigured is a no-op and never spawns git; a non-zero git exit and a throwing lister both return clean without throwing; branch-only markers are extracted.
- Full `@revdev/daemon` suite green (699 tests), workspace build green (`pnpm -r --filter=!studio build`), `pnpm typecheck:studio` green, Biome clean on touched files.

## Notes

No new RPC method; `RPC_METHODS` and the rpc-contract test are untouched. Zero authored regex; explicit return types on exports; no em dashes.

### DIFF
--- a/packages/daemon/src/__tests__/dup-work-claim.test.ts
+++ b/packages/daemon/src/__tests__/dup-work-claim.test.ts
@@ -1,0 +1,167 @@
+/**
+ * dup-work-claim check unit tests.
+ *
+ * Covers the check's invariants:
+ *   - two worktrees claiming the same configured marker -> a warning
+ *   - distinct markers -> clean
+ *   - unconfigured (no markers) -> no-op, and git is never spawned
+ *   - a git failure (non-zero exit OR a throwing lister) -> clean, never throws
+ *
+ * The warning-path tests drive the REAL default lister (hardened `runGit`)
+ * against a throwaway git repo with linked worktrees, so the porcelain parse +
+ * extraction are exercised end to end. The edge cases inject a lister so no git
+ * is spawned and the failure paths are deterministic.
+ *
+ * @vitest-environment node
+ */
+
+import { execFileSync } from 'node:child_process';
+import { mkdir, mkdtemp, rm, writeFile } from 'node:fs/promises';
+import { tmpdir } from 'node:os';
+import { join } from 'node:path';
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
+import { createDupWorkClaimCheck } from '../session-checks/index.js';
+import type { ShellResult } from '../vcs.js';
+
+vi.setConfig({ testTimeout: 20_000, hookTimeout: 20_000 });
+
+const CTX_AGENT = 'agent-test';
+
+function git(cwd: string, args: string[]): void {
+  execFileSync('git', ['-c', 'user.name=t', '-c', 'user.email=t@t', ...args], {
+    cwd,
+    stdio: 'ignore',
+  });
+}
+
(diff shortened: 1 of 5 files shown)


RevealUIStudio:
## Review (fable/opus)

Not a gated surface per the fleet checker. Reviewed as a correctness + safety pass since it hardens a filesystem-reading advisory check and adds a git-reading one.

**Containment hardening (doc-locations) — correct.** resolveWithinRoot resolves both root and the joined path and applies a separator-safe within() (target === root || target.startsWith(root + sep)), so a config dir with a `../` segment is skipped-and-warned instead of read. The separator-safe check correctly avoids the /a/bc-inside-/a/b false positive. Still read-only, warn-only, never throws. This is exactly the nit I flagged on #288, closed properly.

**dup-work-claim — correct and provider-agnostic.** No-op by default (empty claimMarkers or no workDir returns clean, never spawns git), so no tracker's id scheme is baked in. Every failure path returns clean without throwing: a thrown lister, a git error, and a non-zero git exit all yield { ok: true, warnings: [] }. Marker extraction is substring + character-walk, zero regex. Reuses the daemon's hardened runGit via a lazy `await import('../vcs.js')` to avoid the genuine load-time cycle (server.ts imports the registry before its handlers map initializes), which is the extend-before-create-correct choice over a raw spawn that would miss the git hardening/timeout/abort.

**Best-effort invariant preserved:** both checks compose into runSessionChecks, which already isolates throws; these additionally never throw on their own. Registration can still never fail on a check.

**Findings: none.** Verdict: APPROVE. Not gated; merge on green is the owner's call.

PR #298, opened 2026-07-18
test(protocol): add contract test suite for @revdev/protocol

RevealUIStudio:
## Summary

- Adds the first test suite for `packages/protocol` (~636 LOC, previously zero tests). 52 tests across 5 files under `packages/protocol/src/__tests__/`.
- Pins `RPC_METHODS` (non-empty keys/values, key-equals-value self-consistency, no two constant entries sharing the same method-name string).
- Pins base58 encode/decode: alphabet shape, round trips (including leading zero bytes), a known external base58 vector ("Hello World!" -> `2NEpo7TZRRrLZSi2U`), and invalid-character rejection.
- Pins DID formatting/parsing: `isValidAgentId`/`isValidFingerprint` boundary lengths (0, 128, 129 chars), `formatDid`/`parseDid` round trip, and the documented null/throw cases.
- Pins the signature zod schemas (`SignaturePayloadSchema`, `SignatureHeaderSchema`, `SignatureEnvelopeSchema`): valid envelope parses, strict-mode rejects unknown keys, and each DoS-guard bound (nonce length, ts non-negative integer, string max lengths) is exercised at the boundary.
- Adds `packages/protocol/vitest.config.ts` scoping discovery to `src/**/*.test.ts` — without it, vitest also picks up the compiled `dist/__tests__/*.test.js` copies left by `tsc` (protocol's build has no tsup entry-scoping like daemon/bridge), silently double-running every test and risking a stale-dist false pass.
- Wires `pnpm --filter @revdev/protocol test` via a `"test": "vitest run --passWithNoTests"` script, matching the existing `daemon`/`bridge` convention exactly. No CI workflow change needed: `ci.yml`'s `test` job runs `pnpm test` -> `pnpm -r test`, which already picks up any workspace package with a `test` script.
- devDependency added: `vitest@^4.1.10`, the same version/resolution already used by `daemon` and `bridge` (`pnpm install` reused the existing resolution, no new package fetched — 3-line lockfile diff).

## Consumer cross-check

The task asked for a check that the daemon's registered/dispatched RPC methods cover `RPC_METHODS`. That check already exists and currently passes: [`packages/daemon/src/__tests__/rpc-contract.test.ts`](../../blob/test/packages/daemon/src/__tests__/rpc-contract.test.ts) imports the daemon's real handler registry (`listRegisteredMethods()`) with no live server required and asserts exact equality against `RPC_METHODS` in both directions. Studio has an equivalent one-way check (`HARNESS_RPC_MAP` in `apps/studio/src/__tests__/lib/invoke.test.ts`). This PR doesn't duplicate either — duplicating them here would need protocol to depend on its own consumer (daemon), an inverted workspace dependency.

## Prove-red

Verified the uniqueness test actually catches the regression class it targets. Temporarily added a duplicate method-name entry to `RPC_METHODS` locally (`pingDuplicate: 'ping'`), re-ran `pnpm --filter @revdev/protocol test`, confirmed both `every key equals its own value` and `has no two keys sharing the same method name` failed red, then reverted before committing (not part of this diff).

## Test plan

- [x] `pnpm --filter @revdev/protocol test` -> 5 files, 52 tests passed
- [x] `pnpm --filter @revdev/protocol typecheck` -> clean
- [x] `pnpm typecheck` (repo-wide) -> clean
- [x] `pnpm lint` (repo-wide) -> no new findings
- [x] `pnpm install --frozen-lockfile` succeeds against the updated lockfile
- [x] Daemon's `rpc-contract.test.ts`, `did.test.ts`, `agent-identity-crypto.test.ts` re-run clean against the rebuilt protocol dist (no regression from these changes)

### DIFF
--- a/packages/protocol/package.json
+++ b/packages/protocol/package.json
@@ -40,10 +40,12 @@
     "build": "tsc",
     "clean": "rm -rf dist",
     "dev": "tsc --watch",
+    "test": "vitest run --passWithNoTests",
     "typecheck": "tsc --noEmit"
   },
   "devDependencies": {
-    "typescript": "^6.0.3"
+    "typescript": "^6.0.3",
+    "vitest": "^4.1.10"
   },
   "engines": {
     "node": ">=24.13.0"
--- a/packages/protocol/src/__tests__/base58.test.ts
+++ b/packages/protocol/src/__tests__/base58.test.ts
@@ -1,0 +1,69 @@
+import { describe, expect, it } from 'vitest';
+import { BASE58_ALPHABET, base58Decode, base58Encode } from '../base58.js';
+
+describe('BASE58_ALPHABET', () => {
+  it('has exactly 58 characters, all distinct', () => {
+    const chars = BASE58_ALPHABET.split('');
+    expect(chars.length).toBe(58);
+    expect(new Set(chars).size).toBe(58);
+  });
+
+  it('excludes the visually ambiguous characters (0, O, I, l)', () => {
+    for (const ambiguous of ['0', 'O', 'I', 'l']) {
+      expect(BASE58_ALPHABET.includes(ambiguous)).toBe(false);
+    }
+  });
+});
+
+describe('base58Encode', () => {
+  it('encodes an empty byte array as an empty string', () => {
+    expect(base58Encode(new Uint8Array(0))).toBe('');
+  });
+
+  it('encodes a single zero byte as a single leading-zero character', () => {
+    expect(base58Encode(new Uint8Array([0]))).toBe('1');
+  });
+
+  it('matches the well-known "Hello World!" base58 vector', () => {
+    const bytes = new TextEncoder().encode('Hello World!');
+    expect(base58Encode(bytes)).toBe('2NEpo7TZRRrLZSi2U');
+  });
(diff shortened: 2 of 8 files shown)


RevealUIStudio:
## Reviewer confirmation: behavior-preserving, Prove-Red red is expected

Non-author reviewing session. The Prove-Red (TypeScript) failure is the gate's product-source classifier catching the new packages/protocol/vitest.config.ts. Confirmed behavior-preserving: the file only sets test.include to src/**/*.test.ts so vitest stops double-running the stale dist-compiled copies of the same tests; it carries zero product runtime behavior. The 52 added tests are contract-pinning regression guards, which pass against base by design (the contracts they pin already hold); the author prove-red'd the uniqueness test locally by injecting a duplicate method constant (documented in the PR body).

Two notes for the verification-enforcement lane: the gate's failure message names a verify:no-behavior-change label applied by a non-author, but that label does not exist in this repo and scripts/ci/prove-red.mjs never reads labels, so the documented exemption path is unwired; and Prove-Red is not among the required checks (only Private Leak Scan is), so the gate is advisory here. Filing the unwired-exemption finding as a gap.

